PR #2, opened 2019-06-21
DM-20274: Add Alarm class and unit test

r-owen:


python/lsst/ts/watcher/alarm.py
```diff
@@ -0,0 +1,205 @@
...
+        self.acknowledged = False
+        self.acknowledged_by = ""
+        self.enabled = True
+
+        self.timestamp_severity_oldest = 0
+        self.timestamp_severity_newest = 0
+        self.timestamp_max_severity = 0
+        self.timestamp_acknowledged = 0
+
+    def __eq__(self, other):
```
awclemen: I think I read in the XML (Events - line 32) that you say that each alarm should have a unique name - therefore an Alarm object would only have to match on name to be the same object - isn't that correct?   Or am I missing some nuance here...

python/lsst/ts/watcher/alarm.py
```diff
@@ -0,0 +1,205 @@
...
+            explaining what is wrong. Ignored if severity is NONE.
+
+        Returns
+        -------
+        updated : `bool`
+            True if the alarm state changed (i.e. if any fields were modified),
+            False otherwise.
+        """
+        severity = AlarmSeverity(severity)
+        if severity == AlarmSeverity.NONE and self.quiescent:
```
awclemen: quiescent: I had to look this one up ;-)

python/lsst/ts/watcher/alarm.py
```diff
@@ -0,0 +1,205 @@
...
+                self.acknowledged = False
+                self.acknowledged_by = ""
+                self.timestamp_acknowledged = curr_tai
+            self.max_severity = self.severity
+            self.timestamp_max_severity = curr_tai
+
+        self._run_callback()
+        return True
+
+    def acknowledge(self, severity, user):
```
awclemen: in Requirement LTS-897-REQ-0001 - the watcher should publish to SAL when an alarm is acknowledged - I don't see that here- maybe elsewhere?

python/lsst/ts/watcher/alarm.py
```diff
@@ -0,0 +1,205 @@
...
+                      "timestamp_max_severity",
+                      "timestamp_acknowledged"):
+            if getattr(self, field) != getattr(other, field):
+                return False
+        return True
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    def set_severity(self, severity, reason):
```
awclemen: So assuming that if an alarm goes from severity.none to another severity - it has become "active".  LTS-897-REQ-0001 then requires a SAL event to be published if that is the case - maybe this happens elsewhere?

tests/test_alarm.py
```diff
@@ -0,0 +1,402 @@
...
+import copy
+import time
+import unittest
+
+from lsst.ts.idl.enums.Watcher import AlarmSeverity
+from lsst.ts import salobj
+from lsst.ts import watcher
+
+
+class AlarmTestCase(unittest.TestCase):
```
awclemen: Tests look extremely through - good job.

python/lsst/ts/watcher/alarm.py
r-owen: The equality test is primarily intended for unit testing. In many tests I call a method that is expected to change or not change the state on an Alarm. In such cases I want to be able to test this by checking equality against a copy.

python/lsst/ts/watcher/alarm.py
r-owen: That's not a good sign. If you found it confusing then others might also find it confusing. I am open to other suggestions.

python/lsst/ts/watcher/alarm.py
r-owen: The Watcher will publish the alarm event whenever the alarm state changes. When an alarm is acknowledged the alarm's `acknowledged` field is set True and LOVE will see this when the Alarm event is published.

My intent is to have the Watcher model or CSC publish events. I haven't written these yet.

python/lsst/ts/watcher/alarm.py
r-owen: Yes. Again the alarm state will change and the Alarm event will be published, but not by the `Alarm` class.

python/lsst/ts/watcher/alarm.py
awclemen: Could we put a comment here for that - one could almost argue it belongs in the testing class since it is testing only - but you can go either way about this.

python/lsst/ts/watcher/alarm.py
awclemen: Maybe dormant or inactive?

r-owen: `inactive` sounds like the opposite of `active`, which this is not. Also I don't like testing for a negative name as `not inactive` is difficult to parse. Dormant might work. I'll ask on the DM naming channel. I confess I slightly prefer `quiescent` as I think it a bit more accurate.

r-owen: And the winner is...."nominal".

r-owen:
Changes made.

PR #3, opened 2019-07-09
DM-20149: Add the Rule class and supporting classes

r-owen:


overview.txt
```diff
@@ -0,0 +1,56 @@
+Overview (this will be incorporated into the sphinx documentation):
+
+The Watcher is a CSC that monitors SAL components and generates alarm events which are
+intended to be displayed on the LOVE interface. The basic idea is described elsewhere.
```
tribeiro: Maybe it would be a good idea to mention the LTS related documents.

python/lsst/ts/watcher/__init__.py
```diff
@@ -19,5 +19,11 @@
...
-from .alarm import *
-from .version import *  # Generated by sconsUtils
+from .base import *
+from .model import *
+from .rules import *
+
+try:
+    from .version import *
+except ImportError:
+    pass
```
tribeiro: Maybe issue a warning?

r-owen: I will if you think it important, but I would rather not otherwise because I'm not sure it tells the user much (except that they are running from source, which they presumably already know) and it requires importing the warnings module into the lsst.ts.watcher namespace (not a terrible thing to do, but still...).

python/lsst/ts/watcher/rules/enabled.py
```diff
@@ -0,0 +1,88 @@
...
+
+import yaml
+
+from lsst.ts import salobj
+from lsst.ts.watcher import base
+
+
+class Enabled(base.BaseRule):
+    """Monitor the summary state of a CSC.
+
```
tribeiro: I think it would be good to expand the description of this rule a little bit further. Something like;

```Monitor the summary state of a CSC. Issues  `base.AlarmSeverity.SERIOUS` if CSC is in `FAULT` state and `base.AlarmSeverity.WARNING` if CSC is not ENABLED. ```

File: tests/data/config/severalrules.yaml [hunk 4]
```diff
@@ -0,0 +1,13 @@
+disabled_sal_components:
+- ATCamera
+rules:
+- class: Heartbeat
+  configs:
+  - name: ATPtg
```
tribeiro: ATPtg is not outputting heartbeats yet. Probably a good idea to remove it or now.

r-owen: This is for unit tests and will not rely on the real ATPtg running, so it won't harm anything. But there is no point setting a bad example so I will change it.

python/lsst/ts/watcher/rules/enabled.py
r-owen: Good idea. I have updated docs for all rules to describe the alarm severities and provide the rule name.

PR #5, opened 2019-09-18
DM-21264: add the ability to mute alarms

r-owen:


python/lsst/ts/watcher/base/alarm.py
```diff
@@ -107,10 +111,69 @@ def acknowledge(self, severity, user):
...
+        self.muted_by = user
+        self.muted_severity = severity
+        curr_tai = salobj.tai_from_utc(time.time())
+        self.timestamp_unmute = curr_tai + duration
+        self.unmute_task.cancel()
+        self.unmute_task = asyncio.create_task(self.unmute_after(duration=duration))
+        self._run_callback()
+
+    def unmute(self, run_callback=True):
+        """Unmute this alarm.
```
parejkoj: Reading this right, you can mute alarms of a given severity, but you can only unmute "all" muted levels? Is it possible to have two "mute"ings active at a time? E.g. what happens if one user does `mute(500, WARNING, "me")` and another then does `mute(100, SERIOUS, "you")`?

python/lsst/ts/watcher/model.py
```diff
@@ -230,6 +230,37 @@ def get_rules(self, name_regex):
         compiled_re = re.compile(name_regex)
         return [rule for name, rule in self.rules.items() if compiled_re.match(name) is not None]
 
+    def mute_alarm(self, name, duration, severity, user):
+        """Mute one or more alarms for a specified duration.
```
parejkoj: "one ore more alarms for a specified duration at a specific severity"? Because you can mute multiple alarms with a regex, but you can do it for a single severity level.

python/lsst/ts/watcher/model.py
```diff
@@ -230,6 +230,37 @@ def get_rules(self, name_regex):
...
+
+        Parameters
+        ----------
+        name : `str`
+            Alarm name.
+            If a regular expression then acknowledge all matching alarms.
```
parejkoj: As in my other comment, please reword the "Alarm name" portion.

python/lsst/ts/watcher/model.py
```diff
@@ -230,6 +230,37 @@ def get_rules(self, name_regex):
...
+        duration : `float`
+            How long to mute the alarm (sec)
```
parejkoj: period.

python/lsst/ts/watcher/model.py
```diff
@@ -230,6 +230,37 @@ def get_rules(self, name_regex):
...
+        severity : `lsst.ts.idl.enums.Watcher.AlarmSeverity` or `int`
+            Severity to acknowledge. If the severity goes above
+            this level the alarm will unacknowledge itself.
```
parejkoj: unacknowledge, or unmute?

python/lsst/ts/watcher/model.py
```diff
@@ -230,6 +230,37 @@ def get_rules(self, name_regex):
...
+        user : `str`
+            Name of user; used to set acknowledged_by.
```
parejkoj: acknowledged_by, or muted_by?

python/lsst/ts/watcher/model.py
```diff
@@ -230,6 +230,37 @@ def get_rules(self, name_regex):
...
+            rule.alarm.mute(duration=duration, severity=severity, user=user)
+
+    def unmute_alarm(self, name):
+        """Unmute one or more alarms.
+
+        Parameters
+        ----------
+        name : `str`
+            Alarm name. If "all" then unmute all alarms.
+            If a regular expression then acknowledge all matching alarms.
```
parejkoj: "acknowledge all" -> "umuted all"?

python/lsst/ts/watcher/model.py
```diff
@@ -230,6 +230,37 @@ def get_rules(self, name_regex):
...
+        for rule in self.get_rules(name):
+            rule.alarm.mute(duration=duration, severity=severity, user=user)
+
+    def unmute_alarm(self, name):
+        """Unmute one or more alarms.
+
+        Parameters
+        ----------
+        name : `str`
+            Alarm name. If "all" then unmute all alarms.
```
parejkoj: Instead of "all", wouldn't "*." suffice, given that its a regex? Is there special logic for the "all" case? If so, I didn't see it.

tests/base/test_alarm.py
```diff
@@ -19,28 +19,52 @@
...
         self.ncalls += 1
 
-    def alarm_iter(self, name="test.alarm"):
+    def alarm_iter(self, callback, name="test.alarm"):
         """Return an iterator over alarms with all allowed values of
         severity and max_severity.
 
         Parameters
         ----------
+        callback : ``callable`` or `None`
```
parejkoj: no backticks around "callable": https://developer.lsst.io/python/numpydoc.html#callable-types

tests/base/test_alarm.py
```diff
@@ -357,12 +386,100 @@ def test_reset(self):
...
+                with self.assertRaises(ValueError):
+                    alarm.mute(duration=bad_duration, severity=good_severity, user=failed_user)
+                with self.assertRaises(ValueError):
+                    alarm.mute(duration=good_duration, severity=bad_severity, user=failed_user)
+                with self.assertRaises(ValueError):
+                    alarm.mute(duration=bad_duration, severity=bad_severity, user=failed_user)
+                self.assertEqual(alarm, initial_alarm)
+
+                # make sure failures also leave muted alarm state unchanged
+                alarm.mute(duration=good_duration, severity=good_severity, user=good_user)
```
parejkoj: You don't do this check below for *each* of the "bad mutings". I think you've coded `mute` safely enough, but I'd be happier if you did the `assertEqual` after each `assertRaises`.

tests/base/test_alarm.py
```diff
@@ -357,12 +386,100 @@ def test_reset(self):
...
+                # check that unmute on unmuted alarm is a no-op
+                original_alarm = copy.copy(alarm)
+                alarm.unmute()
+                self.assertEqual(alarm, original_alarm)
+                self.assertEqual(self.ncalls, 1)
+
+                # mute alarm and unmute it again before it unmutes itself
+                alarm.mute(duration=duration, severity=severity, user=user)
+                curr_tai = salobj.tai_from_utc(time.time())
+                self.assertEqual(self.ncalls, 2)
```
parejkoj: Ah, now I understand what that thing in `setUp` is doing. That's rather a rather clever use of testing the callback.

tests/base/test_alarm.py
```diff
@@ -357,12 +386,100 @@ def test_reset(self):
...
+                self.assertTrue(alarm.muted)
+                self.assertEqual(alarm.muted_by, user)
+                self.assertEqual(alarm.muted_severity, severity)
+                self.assertGreaterEqual(curr_tai + duration, alarm.timestamp_unmute)
```
parejkoj: GreaterEqual or just Greater? I think this is a test that the alarm unmuted itself "early", but the line below makes me less sure of that.

tests/test_csc.py
```diff
@@ -37,11 +37,17 @@
 
 
 class Harness:
-    def __init__(self, initial_state, config_dir=None):
+    """Make a Watcher CSC and a remote for it.
+
+    Parameters
+    ----------
+    config_dir : `str` (optional)
```
parejkoj: There's also an `initial_state` parameter. I think this is all stuff we should lift into that "how to write CSC tests" doc.

tests/test_csc.py
```diff
@@ -151,6 +156,64 @@ def setUp(self):
...
+            while len(muted_names) < nrules:
+                data = await harness.remote.evt_alarm.next(flush=False, timeout=STD_TIMEOUT)
+                self.assertEqual(data.mutedSeverity, AlarmSeverity.SERIOUS)
+                self.assertEqual(data.mutedBy, user1)
+                muted_names.append(data.name)
+            self.assertEqual(set(muted_names), set(harness.csc.model.rules))
+
+            unmuted_names = []
+            while len(unmuted_names) < nrules:
+                data = await harness.remote.evt_alarm.next(flush=False, timeout=STD_TIMEOUT)
```
parejkoj: I feel like this block, and the one above it, could be turned into a single method "check_mutedness" or something, that takes a severity and a user and does all this checking. Because unless I'm misreading, they're otherwise identical.

tests/test_csc.py
```diff
@@ -151,6 +156,64 @@ def setUp(self):
...
+                                                    severity=AlarmSeverity.SERIOUS,
+                                                    mutedBy=user1,
+                                                    timeout=STD_TIMEOUT)
+            muted_names = []
+            while len(muted_names) < nrules:
+                data = await harness.remote.evt_alarm.next(flush=False, timeout=STD_TIMEOUT)
+                self.assertEqual(data.mutedSeverity, AlarmSeverity.SERIOUS)
+                self.assertEqual(data.mutedBy, user1)
+                muted_names.append(data.name)
+            self.assertEqual(set(muted_names), set(harness.csc.model.rules))
```
parejkoj: Isn't this assert a tautology, given the nature of the while loop above? I don't see how we could exit the while loop and have this assert fail.

tests/test_csc.py
```diff
@@ -151,6 +156,64 @@ def setUp(self):
...
+            await harness.remote.cmd_mute.set_start(name=full_name,
+                                                    duration=5,
+                                                    severity=AlarmSeverity.SERIOUS,
+                                                    mutedBy=user2,
+                                                    timeout=STD_TIMEOUT)
+            data = await harness.remote.evt_alarm.next(flush=False, timeout=STD_TIMEOUT)
+            self.assertEqual(data.name, full_name)
+            self.assertEqual(data.mutedSeverity, AlarmSeverity.SERIOUS)
+            self.assertEqual(data.mutedBy, user2)
+            # that should be the only alarm event from mute
```
parejkoj: "from mute"?

tests/test_model.py
```diff
@@ -280,6 +280,88 @@ def setUp(self):
...
+                self.assertFalse(rule.alarm.muted)
+                self.assertEqual(rule.alarm.muted_severity, AlarmSeverity.NONE)
+                self.assertEqual(rule.alarm.muted_by, "")
+
+            # mute one rule by full name
+            harness.model.mute_alarm(name=full_rule_name,
+                                     duration=5,
+                                     severity=AlarmSeverity.WARNING,
+                                     user=user)
+            for name, rule in harness.model.rules.items():
```
parejkoj: Again, there are two things in the list, so I think unrolling the loop is easier to read and debug.

tests/test_model.py
```diff
@@ -280,6 +280,88 @@ def setUp(self):
...
+                    self.assertEqual(rule.alarm.muted_severity, AlarmSeverity.WARNING)
+                    self.assertEqual(rule.alarm.muted_by, user)
+                else:
+                    self.assertFalse(rule.alarm.muted)
+                    self.assertEqual(rule.alarm.muted_severity, AlarmSeverity.NONE)
+                    self.assertEqual(rule.alarm.muted_by, "")
+
+            # unmute one rule by full name
+            harness.model.unmute_alarm(name=full_rule_name)
+            for rule in harness.model.rules.values():
```
parejkoj: I feel like I've seen this "for rule in rules: all rules are muted" assert loop in quite a few places. Can you lift it into a testing function that can be used in multiple places? e.g. `check_all_rules_muted`

python/lsst/ts/watcher/model.py
```diff
@@ -158,20 +159,21 @@ def disable(self):
...
     def acknowledge_alarm(self, name, severity, user):
-        """Acknowledge the named alarm.
+        """Acknowledge one or more alarms.
 
         Parameters
         ----------
         name : `str`
-            Alarm name
+            Alarm name.
+            If a regular expression then acknowledge all matching alarms.
```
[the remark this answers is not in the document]
r-owen: Good suggestion.

python/lsst/ts/watcher/model.py
```diff
@@ -216,6 +218,18 @@ def add_rule(self, rule):
...
+        """Get all rules whose name matches the specified regular expression.
+
+        Parameters
+        ----------
+        name_regex : `str`
+            Rule/alarm name.
+            If a regular expression then return all matching rules.
+        """
+        compiled_re = re.compile(name_regex)
+        return [rule for name, rule in self.rules.items() if compiled_re.match(name) is not None]
```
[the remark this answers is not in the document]
r-owen: It's a nuisance for the unit test but still probably worth doing. I'll change it to:
```
return (rule for name, rule in self.rules.items() if compiled_re.match(name) is not None)
```

tests/test_csc.py
r-owen: No. The loop picks up *all* alarm events. They don't have to be triggered by muting or unmuting.

What I am testing is that the first batch of events is for the alarms to be muted. Once those have arrived I wait for more and check that they are unmuted.

I'll add a code comment.

tests/test_csc.py
r-owen: How about (and lengthening the timeout):
```
            # There should be the only alarm event from the mute command.
            with self.assertRaises(asyncio.TimeoutError):
                await harness.remote.evt_alarm.next(flush=False, timeout=1)
```
and similarly for the unmute test shortly thereafter

python/lsst/ts/watcher/model.py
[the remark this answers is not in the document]
parejkoj: Should that be `return` or `yield`?

tests/test_csc.py
r-owen: I prefer the current code so I can see exactly what is happening without having to hunt up the definition of another method.

I considered an `assertAlarmEvent` method but it would need a lot of keyword arguments (or a kwargs) and it's a nuisance for boolean fields because I don't want to test those for equality.

tests/test_model.py
r-owen: I prefer it as is. If I add more items to the list the test will still pass.

tests/test_model.py
```diff
@@ -32,6 +31,8 @@
 
 
 class GetRuleClassTestCase(unittest.TestCase):
+    """Test Watcher.get_rule_class.
```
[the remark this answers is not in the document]
r-owen: OK. And I can fix a typo: it's actually `lsst.ts.watcher.get_rule_class`

tests/test_model.py
r-owen: Fair enough. I'll add `assertAlarmMuted` and `assertAlarmNotMuted` methods. That still leaves the loops gets rid of a lot of boilerplate.

tests/test_csc.py
parejkoj: Oh, ok, I see the `set` part of the test now. Your addition there helps, too.

python/lsst/ts/watcher/model.py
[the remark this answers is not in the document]
r-owen: Either would probably work, but it returns a generator expression so I think `return` is clearer

tests/test_model.py
```diff
@@ -53,140 +54,231 @@ def test_bad_names(self):
...
+        self.read_severities = dict()
+        self.read_max_severities = dict()
 
-        self.controller = salobj.Controller(name=name, index=index)
-        self.model = watcher.Model(domain=self.controller.domain,
+        self.controllers = []
+        for name_index in names:
+            name, index = salobj.name_to_name_index(name_index)
+            self.controllers.append(salobj.Controller(name=name, index=index))
+        self.model = watcher.Model(domain=self.controllers[0].domain,
```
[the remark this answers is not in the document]
r-owen: It makes no difference. I just need a domain and I'd rather use an existing one than create a new one.

tests/test_model.py
```diff
@@ -53,140 +54,231 @@ def test_bad_names(self):
...
+            self.read_severities[name] = []
+            self.read_max_severities[name] = []
 
     def alarm_callback(self, alarm):
-        self.read_severities.append(alarm.severity)
-        self.read_max_severities.append(alarm.max_severity)
+        self.read_severities[alarm.name].append(alarm.severity)
+        self.read_max_severities[alarm.name].append(alarm.max_severity)
+        print(f"alarm_callback({alarm.name}, severity={alarm.severity!r}): "
+              f"read_severities={self.read_severities[alarm.name]}")
```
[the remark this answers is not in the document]
r-owen: I find it useful for debugging problems.

tests/test_model.py
```diff
@@ -53,140 +54,231 @@ def test_bad_names(self):
...
+            # acknowledge one rule by full name but not the other
+            harness.model.acknowledge_alarm(name=full_rule_name,
                                             severity=AlarmSeverity.WARNING,
                                             user=user)
-            self.assertTrue(harness.rule.alarm.acknowledged)
-            self.assertEqual(harness.rule.alarm.acknowledged_by, user)
+            for name, rule in harness.model.rules.items():
+                if name == full_rule_name:
+                    self.assertTrue(rule.alarm.acknowledged)
+                    self.assertEqual(rule.alarm.acknowledged_by, user)
```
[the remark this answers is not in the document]
r-owen: I personally feel the test is clearer as written. The rule whose name is full_rule_name is special and the others (whether 1 or more) are not.

tests/test_model.py
```diff
@@ -53,140 +54,231 @@ def test_bad_names(self):
...
+        async with EnabledRulesHarness(names=remote_names) as harness:
+
+            harness.model.enable()
+            await harness.model.enable_task
+
+            self.assertEqual(len(harness.model.rules), nrules)
+
+            for rule in harness.model.rules.values():
+                self.assertTrue(rule.alarm.nominal)
+                self.assertFalse(rule.alarm.acknowledged)
```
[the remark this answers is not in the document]
r-owen: Yes. I use this pattern in all the tests. I've added a code comment `# Check intial conditions.`

python/lsst/ts/watcher/model.py
```diff
@@ -155,6 +155,8 @@ def disable(self):
     async def close(self):
         """Stop rules and close remotes.
         """
+        for rule in self.rules.values():
```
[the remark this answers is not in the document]
r-owen: `self.rules` is a dict of `name: rule` and I want the rules, not the names.

python/lsst/ts/watcher/base/alarm.py
r-owen: An alarm can only be in one muted state. The most recent call wins. In practice it is impossible for an operator to mute an already-muted alarm (unless its current value is higher than muted_severity, at which point it is displayed as a normal alarm), because the alarm has to be displayed in LOVE, the operator interface, and it won't be if it is muted.

r-owen: Also I don't mute an alarm "of a given severity" but "for a given severity". In other words the alarm will be muted as long as its current or max severity is less than or equal to its mute severity.

r-owen: I'll try to clarify this a bit in the doc string for mute. It already said something like this, but I think this is a minor improvement:
```
        severity : `lsst.ts.idl.enums.Watcher.AlarmSeverity` or `int`
            Severity to mute. If the alarm's current or max severity
            goes above this level the alarm should be displayed.
```

r-owen: I will also add this note to `mute`:
```
        Notes
        -----
        An alarm cannot have multiple mute levels and durations.
        If mute is called multiple times, the most recent call
        overwrites information from earlier calls.
```

python/lsst/ts/watcher/model.py
r-owen: I'll say "for a specified duration and severity"

python/lsst/ts/watcher/model.py
r-owen: Yes. I did this everywhere.

python/lsst/ts/watcher/model.py
r-owen: This was a copy/paste error that I have fixed. If the level goes above severity nothing happens to the mute condition of the alarm, but the UI should display the alarm normally.

python/lsst/ts/watcher/model.py
r-owen: Outdated info. I switched to using "*" for the reasons you gave. Fixed.

python/lsst/ts/watcher/model.py
r-owen: Copy/paste error. "all" is not special. Fixed.

python/lsst/ts/watcher/model.py
r-owen: Copy/paste error. Fixed.

tests/base/test_alarm.py
r-owen: This line tests that timestamp_unmute is set plausibly. I agree it should always be > but in case of limited precision for time.time() I didn't want it to needlessly fail. It's a sanity check.

r-owen: In fact the next line is the more interesting line as it checks more carefully.

r-owen: This is a check that the alarm unmutes itself *at all*.

tests/test_csc.py
r-owen: I actually ditched that parameter as it is not practical to start the Watcher in anything other than STANDBY (because it needs a config file).

tests/test_model.py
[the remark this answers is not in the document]
parejkoj: Could you pull that out into a `check_initial_conditions` method, since it shows up everywhere?

tests/base/test_alarm.py
parejkoj: Can you please add comments re: your explanation above, so the rationale here is clear?

tests/test_model.py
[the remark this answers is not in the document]
parejkoj: Please add a comment to that effect (if you haven't already).

PR #9, opened 2019-11-13
DM-22122: add the showAlarms command

r-owen:


python/lsst/ts/watcher/watcher_csc.py
```diff
@@ -113,19 +113,40 @@ def alarm_callback(self, alarm):
...
+        """Show all alarms.
+        """
+        # First get a list of active (not nominal) alarms.
+        # We can iterate over this list, reporting alarm events and yielding
+        # yielding the event loop, without the list changing.
+        # Note: it is not a problem if alarms change state while this command
+        # is running because those state changes will simply trigger
+        # additional alarm events.
+        active_alarms = [rule.alarm for rule in self.model.rules.values() if not rule.alarm.nominal]
+        print(f"showAlarms found these active alarms: {active_alarms}")
```
tribeiro: Any reason for using `print` here and not `log`?

r-owen: Good catch. In fact it was purely for diagnostics while I worked on the code. I removed it.

I also rewrote the code comment. I hope it is clearer now.

PR #19, opened 2020-09-09
DM-26624: Add test of standard state transitions.

r-owen:
This requires ts_salobj 6, due to the new settingsToApply argument
to BaseCscTestCase.check_standard_state_transitions.


wvreeven (approved):
LGTM

PR #24, opened 2020-10-30
DM-26497: update for ts_salobj 6.1.

r-owen:
Note that the previous version is also compatible with ts_salobj 6.1, but the command line parser will lie to you by claiming to support `--state`, which will fail because Watcher needs a non-blank settingsToApply.


mareuter (approved):
Looks good to me.

PR #28, opened 2020-12-03
Tickets/dm27480

rbovill:


r-owen (approved):
A very nice simplification

r-owen:
Please add a new entry to the version history with the 3rd version field incremented. Or I can do that if you prefer. But it'd be nice to use this ticket branch instead of having to create a new one.

PR #30, opened 2021-03-03
DM-28926: store the config schema as code

r-owen:


python/lsst/ts/watcher/config_schema.py
```diff
@@ -1,9 +1,38 @@
+# This file is part of ts_watcher.
+#
+# Developed for the LSST Data Management System.
```
wvreeven, line 3: Please update this and all other files to Vera Rubin Observatory.

python/lsst/ts/watcher/config_schema.py
```diff
@@ -1,9 +1,38 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = ["CONFIG_SCHEMA"]
+
+import yaml
+
+CONFIG_SCHEMA = yaml.safe_load(
+    """
 $schema: http://json-schema.org/draft-07/schema#
 $id: https://github.com/lsst-ts/ts_watcher/Watcher.yaml
```
wvreeven, line 29: Please update the path to reflect the correct file.

setup.py
```diff
@@ -26,7 +26,6 @@
     package_dir={"": "python"},
     packages=setuptools.find_namespace_packages(where="python"),
     package_data={"": ["*.rst", "*.yaml"]},
-    data_files=[(data_files_path / "schema", ["schema/Watcher.yaml"])],
```
wvreeven, line 29: Apparently one can only add comments at or near modified lines. It looks like tools_path, base_prefix and data_files_path are not needed anymore. If that's correct, please remove them and the pathlib import as well.

wvreeven (approved):
Some minor comments but otherwise looks good.

PR #37, opened 2021-09-20
DM-31746: add DewPointDepression rule, plus FilteredTopicWrapper, FilteredFieldWrapper and related classes

r-owen:


python/lsst/ts/watcher/model.py
```diff
@@ -279,6 +307,34 @@ def get_rules(self, name_regex):
             if compiled_re.match(name) is not None
         )
 
+    def make_filtered_topic_wrapper(self, topic, filter_field):
```
pkubanek, line 310: Can you please add an example how to use make_filtered_topic_wrapper? What I am missing is something like - say you would like to issue an alarm (call a method?) when ESS.sensor[1].value > 50. To do that, you need to.. + an example. Would that be possible?

pkubanek (approved):
Looks good. Except for a typo and possible overhaul of JenkinsFile (which I am able to help with and can be done in another ticket), I would like to see a simple example how this can be used to raise alarms. That might be clear from code, but adding that to README.md or make_filtered.. method would help tremendously. Something like - to print "Alarm raised" when ESS.sensor.value > 50, do that:
```python
import ....

model = Model()
model.make_fil...(ESS, "sensor")
model.add_alarm("value", >, 50) # really not sure how I can use that, this will probably not be the case..
```

doc/version_history.rst
```diff
@@ -11,15 +11,32 @@ v1.6.0
...
 
-* `BaseRule` add a default implementation of `BaseRule.is_usable`.
-  Also add an attribute ``remote_keys``, which `BaseRule.is_usable` uses.
-  This default implementation is usable for all existing rules,
-  and it is more efficient than the existing code (except two test rules, for which it is slightly less efficient).
-* `Model`: change the type of disabled_sal_components from list to frozenset.
+* Add rule `rules.DewPointDepression`.
 * Add rule `rules.MTCCWFollowingRototor`: warn when the MT camera cable wrap is not following the camera rotator.
+* Add classes  `FieldWrapperList`, `BaseFilteredFieldWrapper`, `FilteredFieldWrapper`, `IndexedFilteredFieldWrapper`, and `FilteredTopicWrapper`.
+  These allow rules to handle data from CSCs such as the ESS, that publish the the same topic with different data for different subystems.
```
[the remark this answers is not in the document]
r-owen: It is definitely true that we plan to run multiple ESS CSCs with different SAL indices.

But it is also true that one ESS CSC can read more than one instance of a given kind of sensor, e.g. two or more humidity sensors. When that happens, the relevant ESS topic is output for all instances of that kind of sensor, and simply reading the current or most recent value is no longer sufficient to read a particular sensor.

In my opinion it is unacceptable to assume that a particular ESS running at a particular SAL index will only have one humidity sensor, for example. We may add another one later.

This is the problem that filtered topic wrappers solves. We can be sure that a rule that reads ESS data is reading the correct sensor, even if we decide to add more instances of a particular sensor later.

python/lsst/ts/watcher/model.py
r-owen: I'll beef up the docs a bit.

However:
* I plan to add an example of an ESS-based rule to the documentation when I add the humidity rule. That's a nice simple rule (much simpler than the dew point depression rule, because it only reads one kind of sensor).
* We don't yet have a use case for a rule directly creating a FilteredTopicWrapper. Rules using ESS data will use filtered _field_ wrappers (FilteredFieldWrapper and IndexedFilteredFieldWrapper). These field wrappers do contain a topic wrapper, but it's an internal detail.

PR #38, opened 2021-09-28
DM-31933: add a Humidity rule

r-owen:


pkubanek (changes requested):
Hysteresis for humidity rule is great. But from my experience, it's also handy to have option for alarm to remain trrigered some time (minutes) after being crossed. See e.g. https://www.caha.es/observing-mainmenu-148/weatherng for CAHA weather data, showing oscillation in humidity. Those oscillations are sensor depended. The rule shall be for alarm to remain on (alarmed, serious) after same time serious level was crossed. Even if humidity drop below hysteresis, remain in serious for some minutes, so humidity doesn't return above hysteresis amount.

Second, and this I believe deserves change (the above can be ignored): I am missing test for max_data_age. Can you please add this? Create a config with some reasonable small max_data_age, and show that alarm goes to severe after time expires.

tribeiro:
Parachuting here, I just want to express my agreement with Petr. I think in addition to hysteresis it should also have a time delay. Something like, "it needs to remain above the deactivation limit for more than a certain amount of time". I think defaults would be something like 30 minutes.

r-owen:
@tribeiro and @pkubanek The Watcher is already designed to display any alarm until it is acknowledged. Thus a transient condition cannot simply appear and vanish again. Alarm rules don't have to do anything special to make this happen. This gives uniform behavior and simplifies the logic of individual rules.

We can also configure the watcher to auto-ack alarms after a period of time (and even auto-unack them if the alarm is acknowledged but the condition persists). I agree the auto-ack time should be long, if we use it at all.

r-owen:
I agree with @pkubanek's that I should test the max_data_age parameter of the Humidity rule. I will also make sure that is tested for the DewPointDepression rule.

r-owen:
I adding the unit test of max_data_age that you sensibly requested. It exposed a minor misfeature with the way I was handling stale data (the alarm for stale data would appear after twice the configured interval). I fixed that and eliminated needlessly repeating the “no data seen for X seconds” alarm event. I did that for both the dew point depression rule and humidity rule. I also added a description of alarm severity levels and auto ack/auto-unack to the user guide. Please have another look.

doc/index.rst
```diff
@@ -35,6 +35,39 @@ Rules that use a disabled SAL component are not loaded.
...
+.. _lsst.ts.watcher.severity_levels:
+
+Severity Levels
+---------------
+
+Alarms have the following available severity levels:
+
+* CRITICAL: Equipment is in danger; phone calls will be made, or texts or emails sent, if not acknowledged in time (a feature that is planned but not yet implemented).
+* SERIOUS: Something is broken.
+* WARNING: Womething is wrong but we can probably keep operating.
```
pkubanek, line 47: W->S (typo)

pkubanek (approved):
Please see typo. Otherwise looks fine, but please check Jenkins CI.

r-owen:
The test failure is very mysterious. I've never seen that before and I don't see how my changes could have induced it. I'm inclined to leave it for now. If it becomes a problem then I'll look into relaxing the condition. Fundamentally that test should work, if asyncio works as I think it does.

PR #41, opened 2021-11-03
DM-32434: DewPointDepression rule: assume a dewPoint field for all humidity sensors (DM-32251).

r-owen:


wvreeven (approved):
Looks good to me.

PR #44, opened 2022-01-25
DM-33384: make tests more robust

r-owen:


python/lsst/ts/watcher/alarm.py
```diff
@@ -36,6 +39,32 @@ class Alarm:
         Name of alarm. This must be unique among all alarms
         and should be of the form system.[subsystem....]_name
         so that groups of related alarms can be acknowledged.
+
+    Attributes
+    ----------
+    name : str
```
wvreeven, line 45: A friendly suggestion to enclose python types in back ticks here and further down as well.

python/lsst/ts/watcher/alarm.py
```diff
@@ -36,6 +39,32 @@ class Alarm:
...
+        The function receives one argument: this alarm.
+        None, the default, means no callback.
+    auto_acknowledge_delay : float
+        The delay (seconds) after which an alarm will be automatically
+        acknowledged. Never if 0, the default.
+    auto_unacknowledge_delay : float
+        The delay (seconds) after which an alarm will be automatically
+        unacknowleddged. Never if 0, the default.
+    escalate_to : str
+        Who to escalate this alarm to. Do not escalate if blank (the default).
```
wvreeven, line 58: You are using both `, the default.` and `(the default).`. Maybe choose one and stick to that for consistency?

python/lsst/ts/watcher/alarm.py
```diff
@@ -272,6 +303,11 @@ def reset(self):
...
         """Set the severity.
 
+        Call the callback function unless the alarm was nominal
+        and remains nominal.
+        Put the new severity on the severity queue (if it exists),
+        regardless of whether the alarm was nominal.
+
         Parameters
         ----------
         severity : `lsst.ts.idl.enums.Watcher.AlarmSeverity` or `int`
```
wvreeven, line 313: You imported `AlarmSeverity`. Is the whole package needed here then or can do you with only the class name? Further down (line 455) you only use the class name.

python/lsst/ts/watcher/alarm.py
```diff
@@ -405,10 +437,92 @@ def assert_equal(self, other, ignore_attrs=()):
...
+                )
+        if severity != expected_severity:
+            raise AssertionError(
+                f"severity={severity!r} != expected_severity{expected_severity!r}"
+            )
+
+    def init_severity_queue(self):
+        """Initialize the severity queue.
+
+        Call this once before calling `assert_next_severity`.
```
wvreeven, line 513: What happens if you call it more than once? My guess is that you lose any previously stored severity items. Is there a way to check and verify that this hasn't happened before initializing the Queue?

python/lsst/ts/watcher/filtered_topic_wrapper.py
```diff
@@ -118,29 +122,49 @@ def __init__(self, model, topic, filter_field):
...
         """
         self.field_wrappers[field_wrapper.filter_value].append(field_wrapper)
 
     def get_data(self, filter_value):
         """Get the most recently seen data for the given filter_value,
         or None if no data seen.
+
+        Parameters
+        ----------
+        filter_value : any
```
wvreeven, line 151: A friendly suggestion to enclose the python type in back ticks here.

tests/rules/test_enabled.py
```diff
@@ -101,16 +101,8 @@ async def test_call(self):
                 assert len(model.rules) == 1
                 rule_name = f"Enabled.{name}:{index}"
                 rule = model.rules[rule_name]
+                rule.alarm.severity_queue = asyncio.Queue()
```
wvreeven, line 104: Why not call `init_severity_queue` here?

tests/rules/test_clock.py
```diff
@@ -46,7 +44,7 @@ async def aput(self, dt):
         self.data.private_origin = self.salinfo.domain.origin
         setattr(self.data, f"{self.salinfo.name}ID", self.salinfo.index)
         self._writer.write(self.data)
-        await asyncio.sleep(0.001)
+        await asyncio.sleep(0)
```
wvreeven, line 47: I see in many other unit tests that you removed `await asyncio.sleep(0.001)` but here you set the sleep time to zero. What effect does that have?

tests/test_topic_callback.py
```diff
@@ -152,34 +169,28 @@ async def test_add_rule(self):
...
-            # cannot add a rule with the same name as an existing rule
+            # Cannot add rule3 because it has the same name as rule2.
             with pytest.raises(ValueError):
-                topic_callback.add_rule(rule2)
+                topic_callback.add_rule(rule3)
 
-            # modify the rule and try again
-            rule2.alarm.name = rule2.alarm.name + "modified"
-            topic_callback.add_rule(rule2)
+            # Modify the name of rule2 and try again. This should work.
```
wvreeven, line 185: It looks like you're modifying the name of rule3 and not rule2.

wvreeven (approved):
Looks good to me. I have several minor comments, most importantly about the implementation of `init_severity_queue`, that I leave for you evaluate and decide if you want to act on them or not.

python/lsst/ts/watcher/alarm.py
r-owen: Good point. I have gone through and looked for all instances where the back-ticks were missing. I fixed all the instances with a specific type (such as str and bool) and left the squishy ones alone (such as callable and DDS data) since Sphinx cannot usefully link to those.

This package doesn't yet use mypy and I'm a bit reluctant to bring the typing module into the mix yet.

python/lsst/ts/watcher/alarm.py
r-owen: Good question! I did some experiments and here is what I found:
* I had not been linking to the ts_idl docs at all: I added an intersphinx_mapping for ts_idl to conf.py.
* Once I added fixed that, I found that the fully qualified name was required to get a useful link. The short name didn't work. I can see why that might be, but it was a bit disappointing.

python/lsst/ts/watcher/alarm.py
r-owen: Yes it replaces the queue. Not a common use case but I'll add some text.

python/lsst/ts/watcher/filtered_topic_wrapper.py
r-owen: In this case "any" can't make a useful link. I replaced "any" with "scalar", as being more accurate (but still not linkable).

tests/rules/test_enabled.py
r-owen: Good suggestion.

tests/rules/test_clock.py
r-owen: Good catch! In fact it does nothing useful here. Removed.

PR #47, opened 2022-05-02
DM-34620: modernize a subclass of WriteTopic and add Kafka compatibility

r-owen:


Jenkinsfile
```diff
@@ -1,125 +1,142 @@
...
-            // This is a "work around" method
+            // Change ownership of the workspace to Jenkins for clean up.
             withEnv(["HOME=${env.WORKSPACE}"]) {
                 sh 'chown -R 1003:1003 ${HOME}/'
             }
 
-            // The path of xml needed by JUnit is relative to
-            // the workspace.
+            // The path of xml needed by JUnit is relative to the workspace.
             junit 'jenkinsReport/*.xml'
```
pkubanek, line 124: Could be deduced from XML_REPORT_PATH?  Something like junit File($XML_REPORT_PATH).getParentFile().getName() + '/*.xml' ?

pkubanek (approved):
Looks good -minor improvements.

Jenkinsfile
r-owen: This is from our official Jenkinsfile template in tssw_developer_guide. If you have any suggested improvements then it'd be great for T&S if you made them to that package.

I confess I'm not sure why you think it would be better to change this line. It looks nice and simple to me.

PR #49, opened 2022-05-31
DM-34986: build with pyproject.toml

r-owen:


wvreeven (approved):
Looks good to me.

r-owen:
I reworked the entry points and deleted the watcher commander, since I don't think we ever use it. (I tend to use them to control motion controllers).

wvreeven (approved):
Looks good to me.

PR #51, opened 2022-06-29
Adjust ATCameraDewar alarm range for the CCD temperature, and the hys…

patrickingraham:
…teresis value which was too large and was not allowing the alarm to shut off.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -38,9 +38,9 @@ pipeline {
     stages {
         stage ('Update branches of required packages') {
             steps {
-                // When using the docker container, we need to change the HOME path
+                // When using the docker container, we need to change the WHOME path
                 // to WORKSPACE to have the authority to install the packages.
-                withEnv(["HOME=${env.WORKSPACE}"]) {
+                withEnv(["WHOME=${env.WORKSPACE}"]) {
                     sh """
                         source /home/saluser/.setup_dev.sh || echo "Loading env failed; continuing..."
 
@@ -78,7 +78,7 @@ pipeline {
         }
         stage('Run unit tests') {
             steps {
-                withEnv(["HOME=${env.WORKSPACE}"]) {
+                withEnv(["WHOME=${env.WORKSPACE}"]) {
                     sh """
                         source /home/saluser/.setup_dev.sh || echo "Loading env failed; continuing..."
                         setup -r .
@@ -89,7 +89,7 @@ pipeline {
         }
         stage('Build documentation') {
             steps {
-                withEnv(["HOME=${env.WORKSPACE}"]) {
+                withEnv(["WHOME=${env.WORKSPACE}"]) {
                     sh """
                         source /home/saluser/.setup_dev.sh || echo "Loading env failed; continuing..."
                         setup -r .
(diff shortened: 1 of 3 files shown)


craiglagegit:
Are we sure this fixes the problem?  It seems to me the problem is not that the limits were wrong, but that the logic is wrong.  If I read correctly, the old limits were -93 and -95.  The temperature was running at -94.16 and never went outside of these limits, so why were there errors?  I'm going from memory, but I think the error message said, "Temp of -94.16 exceeds limit of -95.0 + 1.0" or something like that.  Why the +1.0?  Is the 1.0 the number that was changed on line 281?  Shouldn't it be -1.0 on the lower side and +1.0 on the upper side?  Or am I confused?  In any case I think we need more verification that this will fix the problem.

patrickingraham:
Sorry, I put this discussion in https://jira.lsstcorp.org/browse/SITCOM-419 but maybe you didn't see it?
See here for why the issue is with the hysteresis value:
https://jira.lsstcorp.org/browse/SITCOM-419?focusedCommentId=724013&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-724013

That being said, we could open the limits a little wider if you want.

craiglagegit:
OK, I read your note, but I'm still lost.  I don't understand what "The reason for the alarm was that it had not surpassed the appropriate amount of "hysteresis" from the alarm limit." means.
The temperature virtually never went outside of (-94.0, -94.35).  So why exactly did it alarm?
________________________________

patrickingraham:
The limit was at -95, with a hysteresis value of 1, meaning it had to go to -94, and stay above that value for several minutes, to clear the alarm. This hysteresis value is used to prevent "fluttering" of alarms when things are near the limit.
However, it was never going to do that because the setpoint was -94.16.

patrickingraham:
no that's the hysteresis value... so that means that if the detector gets too cold, then the temperature will have to get to -95.16+0.75 = -94.51 and then the alarm will clear after 3 minutes.

craiglagegit:
OK, I think I see.  I'll approve it now.  Thanks for explaining.

________________________________

craiglagegit:
OK, I think I understand now.  I think this will fix the problem, so I am approving.

craiglagegit:
I see that I am not a reviewer, but I agree with the changes, FWIW.

Jenkinsfile
```diff
@@ -116,8 +116,8 @@ pipeline {
     post {
         always {
             // Change ownership of the workspace to Jenkins for clean up.
-            withEnv(["HOME=${env.WORKSPACE}"]) {
-                sh 'chown -R 1003:1003 ${HOME}/'
+            withEnv(["WHOME=${env.WORKSPACE}"]) {
+                sh 'chown -R 1003:1003 ${WHOME}/'
```
r-owen, line 120: I believe this should be left as `HOME` (only in this final cleanup section). That is certainly what I was told and what I have done in my other packages (as I get to them).

r-owen (changes requested):
Thank you for updating the Jenkinsfile, but I think the changes in the post section may need to be undone (though I'd be thrilled if that was not the case -- it's easier to use WHOME everywhere).

The changes to the default values of the ATCameraDewar rule look good to me.

r-owen:
> OK, I think I understand. But looking at the pull request, I don't see the change to the hysteresis value. Is it somewhere else? I see line 281, where "The amount by which a value must improve past an alarm threshold before the alarm severity decreases." is changed from 1.0 to 0.75. Is that something else?
> […](#)

If you expand the text a bit in the diff you will see that the parameter you mentioned is named "temperature_hysteresis".

Jenkinsfile
patrickingraham: OK... I don't really understand why this is but that's ok. I thought it was a global change.
Thanks for the catch.

r-owen (approved):

PR #54, opened 2022-10-05
DM-36434: switch from OpsGenie to PagerDuty

r-owen:


python/lsst/ts/watcher/config_schema.py
```diff
@@ -105,26 +105,9 @@
...
-                description: ->
-                    Name of team, user, or schedule to which to escalate this alarm.
-                    Note that IDs are not supported; it must be a name.
-                type: string
-              type:
-                description: Type of name.
-                type: string
-                enum: ["team", "user", "schedule"]
+        responder:
+          description: Who to escalate this alarm to.
```
wvreeven, line 109: In `alarm.py` the description is `Who or what to escalate the alarm to.` so maybe use that here as well for consistency?

python/lsst/ts/watcher/mock_pager_duty.py
```diff
@@ -0,0 +1,239 @@
...
+import types
+import typing
+import uuid
+from http import HTTPStatus
+
+from aiohttp import web
+
+
+def get_event_key() -> str:
+    """Return a new unique request key"""
```
wvreeven, line 35: Please end the sentence in a `.`.

python/lsst/ts/watcher/mock_pager_duty.py
```diff
@@ -0,0 +1,239 @@
...
+            "component",
+            "group",
+            "class",
+            "custom_details",
+        )
+    )
+
+    def __init__(self, port: int = 80) -> None:
+        self.port = port
+        self.escalation_key = os.environ["ESCALATION_KEY"]
```
wvreeven, line 99: What happens if this env var doesn't exist? Would it make sense to add error handling here?

python/lsst/ts/watcher/mock_pager_duty.py
```diff
@@ -0,0 +1,239 @@
...
+        # Dict of event ID: data_dict
+        self.events: dict[str, dict] = dict()
```
wvreeven, line 101: Can you be more explicit about `data_dict`? What does that mean? Any example or a URL that points to one?

python/lsst/ts/watcher/mock_pager_duty.py
```diff
@@ -0,0 +1,239 @@
...
+
+    Attributes
+    ----------
+    port : `int`
+        If the specified port is 0, then the chosen port
+        (after `start` has run). Otherwise the specified port.
+        The default is 80 because that is what PagerDuty uses.
+    url : `str`
+        The root URL of the service. "" until `start` has run.
+    """
```
wvreeven, line 69: It looks like MockPagerDuty contains more attributes (escalation_key, events, runner, etc). Can you add those here as well please?

python/lsst/ts/watcher/mock_pager_duty.py
```diff
@@ -0,0 +1,239 @@
...
+        self.url: str = ""
+        # Unit tests may set this True to reject the next request
+        self.reject_next_request = False
+
+    def make_app(self) -> web.Application:
+        """Make an instance of the web application."""
+        app = web.Application()
+        app.add_routes(
+            [
+                # web.post("/:{identifier}/close", self.close_event),
```
wvreeven, line 113: Are this line and line 115 required to stay? If yes, please add a comment explaining that and otherwise please remove them.

python/lsst/ts/watcher/mock_pager_duty.py
```diff
@@ -0,0 +1,239 @@
...
+
+    async def start(self) -> None:
+        """Start the service.
+
+        Raises
+        ------
+        RuntimeError
+            If port = 0 and serving on more than one socket
+            (in which case the served port cannot be determined).
+        """
```
wvreeven, line 128: A RuntimeError also is raised if self.runner is not None. Please add that to the doc string.

python/lsst/ts/watcher/mock_pager_duty.py
```diff
@@ -0,0 +1,239 @@
...
+        if self.port == 0:
+            if len(server.sockets) != 1:
+                raise RuntimeError(
+                    "Serving on more than one socket; cannot determine the port"
+                )
+            self.port = server.sockets[0].getsockname()[1]  # type: ignore
+        self.url = f"http://127.0.0.1:{self.port}"
+
+    async def close(self) -> None:
+        """Stop the service."""
```
wvreeven, line 146: Please add a Raises section for the RuntimeError.

python/lsst/ts/watcher/mock_pager_duty.py
```diff
@@ -0,0 +1,239 @@
...
+        Parameters
+        ----------
+        routing_key : `str`
+            Routing key.
+
+        Raises
+        ------
+        HTTPInternalServerError
+            If self.reject_next_request True.
+        Raise HTTPForbidden
```
wvreeven, line 165: I think you can remove `Raise ` here. Also, HTTPBadRequest is raised and not HTTPForbidden.

python/lsst/ts/watcher/mock_squadcast.py
```diff
@@ -0,0 +1,229 @@
...
+    port : `int`
+        If the specified port is 0, then the chosen port
+        (after `start` has run). Otherwise the specified port.
+        The default is 80 because that is what OpsGenie uses.
+    url : `str`
+        The root URL of the service. "" until `start` has run.
+    endpoint_url : `str`
+        The URL of the endpoint (including the secret API key).
+        "" until `start` has run.
+    """
```
wvreeven, line 109: The MockSquadCast class contains more attributes. Please add them here.

python/lsst/ts/watcher/mock_squadcast.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def __init__(self, port: int = 80) -> None:
+        self.port = port
+        self.escalation_key = os.environ["ESCALATION_KEY"]
+        # Dict of alert ID: data_dict
```
wvreeven, line 114: Again, what is a `data_dict`?

python/lsst/ts/watcher/mock_squadcast.py
```diff
@@ -0,0 +1,229 @@
...
+
+    async def start(self) -> None:
+        """Start the service.
+
+        Raises
+        ------
+        RuntimeError
+            If port = 0 and serving on more than one socket
+            (in which case the served port cannot be determined).
+        """
```
wvreeven, line 138: A RuntimeError is raised also when already started.

python/lsst/ts/watcher/mock_squadcast.py
```diff
@@ -0,0 +1,229 @@
...
+        Parameters
+        ----------
+        request : `aiohttp.web.Request`
+            Request.
+
+        Raises
+        ------
+        `web.HTTPRequestEntityTooLarge`
+            If the request is too large.
+        `web.HTTPUnauthorized`
```
wvreeven, line 174: Several HTTPBadRequests get raised but not HTTPUnauthorized. Please amend the doc string accordingly.

python/lsst/ts/watcher/watcher_csc.py
```diff
@@ -139,48 +146,50 @@ async def escalate_alarm(self, alarm):
...
+            * alarm.escalation_responder empty: there is nobody to escalate
               the alarm to (so do_escalate should never have been set).
         """
         if alarm.escalated_id:
             raise RuntimeError("Alarm already escalated")
         if not alarm.do_escalate:
             raise RuntimeError("Alarm do_escalate false")
-        if not alarm.escalation_responders:
-            raise RuntimeError("Alarm escalation_responders empty")
+        if not alarm.escalation_responder:
```
wvreeven, line 156: Does `not` work correctly on an empty string as well? Because I seem to have understood that that may get passed on and not `None`.

wvreeven (changes requested):
I have a lot of minor comments and one worry for a potential bug in the Watcher CSC. You use a `not` to validate a string and I think that doesn't work. Please push back if I am wrong.

python/lsst/ts/watcher/mock_squadcast.py
r-owen: Thanks! That was a holdover from when I thought an authorization header was required (the vendor's docs are not very thorough). It definitely needs to be deleted.

python/lsst/ts/watcher/watcher_csc.py
r-owen: An empty string is treated as logically false. And this handles None as well as "" (I don't remember if None is an allowed value, but I can imagine that in a future code update it might be one).

python/lsst/ts/watcher/mock_pager_duty.py
r-owen: OK. I'll raise a clearer error and document it in all three mock services.

python/lsst/ts/watcher/mock_pager_duty.py
r-owen: I will remove them. The PagerDuty service supports them but I am not sure if I would need them. Since we're not planning to use PagerDuty yet, I won't learn this in time to do anything about it.

python/lsst/ts/watcher/mock_pager_duty.py
r-owen: Good point. I decide to omit the raise and silently work if already closed (for all three mock services).

python/lsst/ts/watcher/mock_pager_duty.py
r-owen: Good catch. And I changed the code to raise HTTPForbidden

wvreeven (approved):
Thanks for the additional changes. Looks really good now!

PR #55, opened 2022-10-13
Tickets/dm 36455

tribeiro:


tests/rules/test_script_failed.py
```diff
@@ -0,0 +1,168 @@
...
+                    ),
+                    (
+                        True,
+                        False,
+                        1002,
+                        ScriptProcessState.DONE,
+                        ScriptState.FAILED,
+                        AlarmSeverity.NONE,
+                        AlarmSeverity.WARNING,
+                    ),
```
r-owen, line 142: This works, but I find it difficult to determine which parameters are set to which values. Consider making a dataclass to hold these test parameters and making a list of those. It adds some duplication of the parameter names, but makes it clearer which values are set to what:
```
@dataclasses.dataclass
class TestParams:
    queue_enabled: bool
    ...
    expected_severity_1: AlarmSeverity

for params in (
    TestParams(
         queue_enabled=True,
         queue_running=True,
         ...,
         expected_severity_2=AlarmSeverity.NONE
     ),
    ...
):
```

tests/rules/test_script_failed.py
```diff
@@ -0,0 +1,168 @@
...
+                    )
+
+                    await script_queue.evt_script.set_write(
+                        scriptSalIndex=script_sal_index,
+                        processState=process_state,
+                        scriptState=script_state,
+                    )
+
+                    # It will publish 2 alarms one for each event published
+                    # by the ScriptQueue.
```
r-owen, line 158: I suggest clarifying this comment. I believe this will publish 2 alarm states for each set of test parameters: one for the queue ScriptQueue event, the next for the script ScriptQueue event. (I read your current text to mean we get 2 alarm states per single ScriptQueue event, which surely is not right).

r-owen (approved):
Very nice indeed.

I have a few very minor suggestions I hope you will consider before merging. A possibly larger issue is whether this rule should report the problem as a WARNING or as SERIOUS?

tests/rules/test_script_failed.py
r-owen: Another option, which would allow you to publish a script state of FAILED for the non-current script (which your rule should ignore) is to just publish the events. More duplication, but no issue of 2 events per test data. I admit that only the current script should ever be reported as FAILED (except if it fails configuration, which is a different state in the soon-to-be-released version).

Or make your test params be for one event or the other, and publish just the one event for each set of parameters.

PR #57, opened 2022-10-27
DM-33898: OCPS calibrations script: add Watcher alarms/warnings when verification fails

plazas:


python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
```diff
@@ -0,0 +1,255 @@
...
+
+import yaml
+import json
+import lsst.daf.butler as dafButler
+
+from lsst.ts.idl.enums.Watcher import AlarmSeverity
+from lsst.ts import watcher
+
+
+class CpVerifyBiasAlarm(watcher.BaseRule):
```
tribeiro, line 32: Instead of `CpVerifyBiasAlarm`, let's make it a generic `CpVerifyAlarm`

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
```diff
@@ -0,0 +1,255 @@
...
+    """Report if cp_verify BIAS tests failed.
```
tribeiro, line 33: > Monitor state of cp_verify executions from the OCPS for a particular type of calibration (BIAS, DARK, FLAT, etc).

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
```diff
@@ -0,0 +1,255 @@
...
+    and issue a WARNING if not.
+
+    Parameters
+    ----------
+    config : `types.SimpleNamespace`
+        Rule configuration, as validated by the schema.
+
+    Notes
+    -----
+    The alarm name is f"CpVerifyBiasAlarm.{name}:{index}",
```
tribeiro, line 45: > The alarm name is f"CpVerifyAlarm.{ocps_index}:{calibration_type}",

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
```diff
@@ -0,0 +1,255 @@
...
+    where name and index are derived from ``config.name``.
```
tribeiro, line 46: > where ``ocps_index`` and ``calibration_type`` are derived from ``config.ocps_index`` and ``config.calibration_type``, respectively.

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
```diff
@@ -0,0 +1,255 @@
...
+    """
+
+    def __init__(self, config):
+        remote_name = "OCPS"
+        # remote_index = 0: process events from both OCPS's
+        remote_index = 0
```
tribeiro, line 52: read this from ``config.ocps_index``

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
```diff
@@ -0,0 +1,255 @@
...
+        remote_info = watcher.RemoteInfo(
+            name=remote_name,
+            index=remote_index,
+            callback_names=["job_result"],
+            poll_names=[],
+        )
+        super().__init__(
+            config=config,
+            name=f"CpVerifyBiasAlarm.{remote_info.name}:{remote_info.index}",
```
tribeiro, line 61: Update name to match the one I proposed above: ``CpVerifyAlarm.{ocps_index}:{calibration_type}``

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
```diff
@@ -0,0 +1,255 @@
...
+            remote_info_list=[remote_info],
+        )
+
```
tribeiro, line 64: you probably want to store in a class variable the type of calibration, e.g.; 

```
self.calibration_type = config.calibration_type
```

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
```diff
@@ -0,0 +1,255 @@
...
+    @classmethod
+    def get_schema(cls):
+        schema_yaml = """
+            $schema: http://json-schema.org/draft-07/schema#
+            description: Configuration for CpVerifyBiasAlarm
+            type: object
+            properties:
+                name:
```
tribeiro, line 72: replace name with `calibration_type`?

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
```diff
@@ -0,0 +1,255 @@
...
+        )
+        super().__init__(
+            config=config,
+            name=f"CpVerifyBiasAlarm.{remote_info.name}:{remote_info.index}",
+            remote_info_list=[remote_info],
+        )
+
+    @classmethod
+    def get_schema(cls):
+        schema_yaml = """
```
tribeiro, line 67: You need to add the OCPS index to the configuration parameter and the calibration_type... you can remove the "name" parameter..

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
```diff
@@ -0,0 +1,255 @@
...
+            $schema: http://json-schema.org/draft-07/schema#
+            description: Configuration for CpVerifyBiasAlarm
+            type: object
+            properties:
+                name:
+                    description: >-
+                        CSC name and index in the form `name` or `name:index`.
+                        The default index is 0.
+                    type: string
+                number_verification_tests_threshold_bias:
```
tribeiro, line 77: maybe just `verification_threshold`?

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
```diff
@@ -0,0 +1,255 @@
...
+                    type: integer
+                    descriptor: Minimum number of verification tests per detector per exposure per \
```
tribeiro, line 79:
```
descriptor: Minimum number of verification tests per detector per exposure per \
                    test type that should pass to certify the calibration.
```

This seems a bit confusing. Is this the number of tests that need to pass or the number of tests that can fail? I think it would be easier to encode how many tests can fail than how many tests can pass... This way, if you increase the number of tests, the verification will still be relevant...

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
```diff
@@ -0,0 +1,255 @@
...
+            additionalProperties: false
+        """
+        return yaml.safe_load(schema_yaml)
+
+    def __call__(self, topic_callback):
+        msg = topic_callback.get()
+        # OCPS response after calling the pipetask
+        response = json.loads(msg.result)
+        # boolean: did bias verification fail?
+        verify_bias_tests_pass = await self.get_verify_tests_pass_boolean(response)
```
tribeiro, line 92: please remove all references to "bias"

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
```diff
@@ -0,0 +1,255 @@
...
+
+        if verify_bias_tests_pass:
+            return watcher.NoneNoReason
+        else:
+            # Tests did not pass. Follow up on images and calibrations is
+            # recommended.
+            return AlarmSeverity.SERIOUS, "FAULT state"
```
tribeiro, line 99: In the docstring above you said it was supposed to be "WARNING" alarm but here you set it so SERIOUS.

tribeiro (commented):
Here is an initial round of comments. Can you turn this into a draft PR while we work to rough details?

One thing I noticed is that we only have 1 level of alarm. Consider having different levels for WARNING, SERIOUS, etc..

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
plazas: Sounds good. I had it like that initially, and I believe we could make this a general alarm.

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
plazas: I kept the long name because that's the same one that it has in the calibration script. In that script, there's also one of these parameters per calibration type: bias, dark, flat.

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
plazas: This is the maximum number of failures per detector per test type.  in the code, we define the max number of failures tolerated per exposure, and then we count the number of exposures that fail this. If this number is more than half the exposures, we fail verification. This is in ``count_failed_verification_tests``.

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
plazas: So, in order to have the same as the calibration script, should we have three parameters here?

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
plazas: I adjusted the docstring to say ``SERIOUS``.

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
tribeiro: I think maybe `ocps_index` should come first...

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
tribeiro: no, use the short version and it will be the one for whatever calibration the alarm is configured for...

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,266 @@
...
+        response = json.loads(msg.result)
+        # boolean: did verification fail?
+        verify_tests_pass = await self.get_verify_tests_pass_boolean(response)
+
+        if verify_tests_pass:
+            return watcher.NoneNoReason
+        else:
+            # Tests did not pass. Follow up on images and calibrations is
+            # recommended.
+            return AlarmSeverity.SERIOUS, "FAULT state"
```
tribeiro, line 102: The string that you pass after the alarm severity appears in the alarm. You could add information about how many tests failed and what was the threshold. Something like:

```
return AlarmSeverity.SERIOUS, f"cp_verify run {job_id} failed verification threshold. Got {n_failures}, threshold {self.verification_threshold}."
```

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,266 @@
...
+            additionalProperties: false
+        """
+        return yaml.safe_load(schema_yaml)
+
+    def __call__(self, topic_callback):
+        msg = topic_callback.get()
+        # OCPS response after calling the pipetask
+        response = json.loads(msg.result)
+        # boolean: did verification fail?
+        verify_tests_pass = await self.get_verify_tests_pass_boolean(response)
```
tribeiro, line 95: I think you need more information out of the `get_verify_test_pass_boolean` than a single boolean. Also, usually it is not a good practice to add the "return value"  to the name of the function.

I think you can return the value itself instead.. so basically this would become:

```
return await self.check_response(response)
```

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,266 @@
...
+
+        if verify_tests_pass:
+            return watcher.NoneNoReason
+        else:
+            # Tests did not pass. Follow up on images and calibrations is
+            # recommended.
+            return AlarmSeverity.SERIOUS, "FAULT state"
+
+    async def get_verify_tests_pass_boolean(self, response_verify):
```
tribeiro, line 104: I think you could rename this `check_response` and make it return the alarm severity + reason.

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,266 @@
...
+
+        Returns
+        -------
+        verify_pass : `bool`
+            Did the cp_verify tests pass?
+        """
+        verify_pass = True
+        # Loop over the entries of the 'results' list
+        # and look for the adecuate dataset type.
+        for entry in response_verify["results"]:
```
tribeiro, line 120: So, the tricky part here is that if you return an ok here, it means an unrelated OCPS run will reset the alarm level. But I think the only thing that would reset the alarm level is another OCPS run that passes verification. 

I think what you will need to know is keep track of the current alarm level and if an OCPS runs comes through that is not intended for this alarm, you just return the current level/reason.. 

Another alternative would be to raise an exception when this occurs, but this means we will  get an exception every time someone runs an unrelated OCPS job. Not a big deal but maybe not ideal as well..

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,266 @@
...
+            if "verifyBiasStats" in entry["uri"]:
+                verify_stats_string = "verifyBiasStats"
+                break
+            elif "verifyDarkStats" in entry["uri"]:
+                verify_stats_string = "verifyDarkStats"
+                break
+            elif "verifyFlatStats" in entry["uri"]:
+                verify_stats_string = "verifyFlatStats"
+                break
+            else:
```
tribeiro, line 130: How many entries are there in `results` that you need to process? This `else` statement will cause the loop to break in the first pass. If there are more than one entries you need to process you probably need to think about a better way to do this.

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,266 @@
...
+                # This is not a response from cp_verify
+                # bias, dark, or flat.
+                # Do not issue a watcher alarm.
+                return verify_pass
+
+        # Find repo and instrument
+        for entry in response_verify["parameters"]["environment"]:
+            if entry["name"] == "BUTLER_REPO":
+                repo = entry["value"]
```
tribeiro, line 139: `repo` is not defined outside this `if` statement, so if this never runs you will end up with undefined `repo`... 

You probably want to check for this condition.

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
```diff
@@ -0,0 +1,255 @@
+# This file is part of ts_watcher.
```
tribeiro, line 1: I think you can remove this one.

tribeiro (commented):
A couple more comments..

python/lsst/ts/watcher/rules/cp_verify_alarm.py
plazas: We know that the OCPS response contains this entry, but if not, I will raise an ``RuntimeError`` (we can't do anything if the response, for whatever reason, does not have a repo).

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,294 @@
...
+        msg = topic_callback.get()
+        # OCPS response after calling the pipetask
+        response = json.loads(msg.result)
+        # Get the dictionary with cp_verify stats, from the butler.
+        verify_stats = await self.get_cp_verify_stats(response)
+        # boolean: did verification fail?
+        return await self.check_response(response, verify_stats)
+
+    async def check_response(self, response_verify, verify_stats):
+        """Determine if cp_verify bias tests passed from OCPS response.
```
tribeiro, line 101: remove "bias"

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,294 @@
...
+        verify_pass = True
+        job_id_verify = response_verify["job_id"]
+
+        if verify_stats["SUCCESS"] is False:
+            (verify_pass, _, thresholds,) = await self.count_failed_verification_tests(
+                verify_stats, self.config.verification_threshold
+            )
+        else:
+            # Nothing failed
+            verify_pass = True
```
tribeiro, line 127: this was already set in line 118, you can probably remove the else statement

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,294 @@
...
+
+        if verify_pass:
+            return watcher.NoneNoReason
+        else:
+            n_exp_failed = thresholds["FINAL_NUMBER_OF_FAILED_EXPOSURES"]
+            n_exp_threshold = thresholds["MAX_FAILED_EXPOSURES_THRESHOLD"]
+            return AlarmSeverity.SERIOUS, (
```
tribeiro, line 134: Let's use WARNING for now

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,294 @@
...
+                verify_stats_string = "verifyDarkStats"
+                break
+            elif "verifyFlatStats" in entry["uri"]:
+                verify_stats_string = "verifyFlatStats"
+                break
+            else:
+                # This is not a response from cp_verify
+                # bias, dark, or flat.
+                raise salobj.ExpectedError(
+                    "Job {job_id_verify} is not a recognizable cp_verify run."
```
tribeiro, line 277: missing an "f" in the string...

```
f"Job {job_id_verify} is not a recognizable cp_verify run."
```

tests/rules/test_cp_verify_alarm.py
```diff
@@ -0,0 +1,376 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+# import asyncio
```
tribeiro, line 22: please, remove commented code lines

python/lsst/ts/watcher/rules/cp_verify_bias_alarm.py
tribeiro: I think we should WARNING instead

python/lsst/ts/watcher/rules/cp_verify_alarm.py
tribeiro: what if `response_verify["parameters"]["environment"]` is empty?

tribeiro:
There's a syntax error in you unit test file:

```
E     File "/home/jenkins/workspace/escope-and-Site_ts_watcher_PR-57/tests/rules/test_cp_verify_alarm.py", line 132
E       2022110300072: {'FAILURES': ['RXX_S00 C00 NOISE', 'RXX_S00 C00 PROCESSING', 'RXX_S00 C01
E                                                                                   ^
E   SyntaxError: unterminated string literal (detected at line 132)
```

plazas:
> There's a syntax error in you unit test file:
> 
> ```
> E     File "/home/jenkins/workspace/escope-and-Site_ts_watcher_PR-57/tests/rules/test_cp_verify_alarm.py", line 132
> E       2022110300072: {'FAILURES': ['RXX_S00 C00 NOISE', 'RXX_S00 C00 PROCESSING', 'RXX_S00 C01
> E                                                                                   ^
> E   SyntaxError: unterminated string literal (detected at line 132)
> ```

Thanks. I now read the example strings from a file.

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,291 @@
...
+from lsst.ts.idl.enums.Watcher import AlarmSeverity
+from lsst.ts import watcher
+
+
+class CpVerifyAlarm(watcher.BaseRule):
+    """Monitor state of cp_verify executions from the OCPS
+    for a particular type of calibration (BIAS, DARK, FLAT, etc).
+
+    Set alarm severity NONE if the cp_verify tests passed,
+    and issue a SERIOUS alarm if not.
```
tribeiro, line 38: Please! Let's change this to WARNING

plazas: I'm so sorry, I changed the code but missed the docstring. I'll change it now.

python/lsst/ts/watcher/rules/cp_verify_alarm.py
plazas: I added a check for its length, and issue an error if is empty.

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,296 @@
...
+    def get_schema(cls):
+        schema_yaml = """
+            $schema: http://json-schema.org/draft-07/schema#
+            description: Configuration for CpVerifyBiasAlarm
+            type: object
+            properties:
+                calibration_type:
+                    description: >-
+                        Alarm calibration type: BIAS, DARK, or FLAT.
+                    type: string
```
r-owen, line 79: If these are the only allowed values then please update the schema to only allow those values. Best to catch errors in the validation step.

I gently suggest making them lowercase to improve readability. You can cast them to uppercase internally when using them. But if all usage on the mountain is uppercase then it's fine to leave them as is.

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,296 @@
...
+                ocps_index:
+                    OCPS index (e.g., OCPS:1: LATISS; OCPS:2: LSSTComCam).
```
r-owen, line 81: If the number of allowed values is fixed and fairly small, please specify them as an enum to catch errors in the validation step.

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,296 @@
...
+            additionalProperties: false
+        """
+        return yaml.safe_load(schema_yaml)
+
+    def __call__(self, topic_callback):
+        msg = topic_callback.get()
+        # OCPS response after calling the pipetask
+        response = json.loads(msg.result)
+        # Get the dictionary with cp_verify stats, from the butler.
+        verify_stats = self.get_cp_verify_stats(response)
```
r-owen, line 96: This call may run slowly and block the Watcher from doing anything else. Please run it in a thread, e.g. using `run_in_executor`. This salobj documentation for writing scripts shows how: https://ts-salobj.lsst.io/sal_scripts.html#run-method

You can also create a concurrent.futures.ThreadPoolExecutor and pass that as the first argument to run_in_executor, instead of None. I suggest that, because you'll be calling this a lot and it is more efficient to create one and keep using it.

Which brings up a question: how often will this run? Will it have enough time to process one message before the next one comes in? In other words: how many butlers will be running queries at the same time? Just curious. if you create your own thread pool executor you can limit this (e.g. if you specify n=1 then only one will run at a time, which could make processing your rule back up, but will not block the Watcher itself, n=3 would allow up to 3 to run at the same time...).

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,296 @@
...
+            if entry["name"] == "BUTLER_REPO":
+                repo = entry["value"]
+                instrument_name = repo.split("/")[-1]
+                break
+            else:
+                raise salobj.ExpectedError("OCPS response does not list a repo.")
+
+        # Collection name containing the verification outputs.
+        verify_collection = f"u/ocps/{job_id_verify}"
+        butler = dafButler.Butler(repo, collections=[verify_collection])
```
r-owen, line 293: If creating butlers is still expensive, please cache the ones you create. Specifically I suggest:
* For the cache use a dict of key: butler, where the key is `(repo, verify_collection)` or `(repo, job_id_verify)`, as you prefer. Together those seem to uniquely identify a butler.
* Create the empty cache in the `__init__` method.
* Empty the cache (and close all butlers) in the `stop` method. You probably know this, but this is a nice way to clear a dict:

```
while mydict:
    key, value = mydict.popitem()
    # ... do something with value, such as close it
```

tests/rules/data/cp_verify_alarm/cp_verify_bias_stats.yaml
```diff
@@ -0,0 +1 @@
...6 PROCESSING', 'RXX_S00 C17 CR_NOISE', 'RXX_S00 C17 NOISE', 'RXX_S00 C17 PROCESSING']}, 2022110400020: {'FAILURES': ['RXX_S00 C00 NOISE', 'RXX_S00 C00 PROCESSING', 'RXX_S00 C01 CR_NOISE', 'RXX_S00 C01 NOISE', 'RXX_S00 C01 PROCESSING', 'RXX_S00 C02 CR_NOISE', 'RXX_S00 C02 NOISE', 'RXX_S00 C02 PROCESSING', 'RXX_S00 C03 CR_NOISE', 'RXX_S00 C03 NOISE', 'RXX_S00 C03 PROCESSING', 'RXX_S00 C04 CR_NOISE', 'RXX_S00 C04 NOISE', 'RXX_S00 C04 PROCESSING', 'RXX_S00 C05 NOISE', 'RXX_S00 C05 PROCESSING', 'RXX_S00 C06 NOISE', 'RXX_S00 C06 PROCESSING', 'RXX_S00 C07 CR_NOISE', 'RXX_S00 C07 NOISE', 'RXX_S00 C07 PROCESSING', 'RXX_S00 C10 CR_NOISE', 'RXX_S00 C10 NOISE', 'RXX_S00 C10 PROCESSING', 'RXX_S00 C11 CR_NOISE', 'RXX_S00 C11 NOISE', 'RXX_S00 C11 PROCESSING', 'RXX_S00 C12 CR_NOISE', 'RXX_S00 C12 NOISE', 'RXX_S00 C12 PROCESSING', 'RXX_S00 C13 CR_NOISE', 'RXX_S00 C13 NOISE', 'RXX_S00 C13 PROCESSING', 'RXX_S00 C14 CR_NOISE', 'RXX_S00 C14 NOISE', 'RXX_S00 C14 PROCESSING', 'RXX_S00 C15 CR_NOISE', 'RXX_S00 C15 NOISE', 'RXX_S00 C15 PROCESSING', 'RXX_S00 C16 CR_NOISE', 'RXX_S00 C16 NOISE', 'RXX_S00 C16 PROCESSING', 'RXX_S00 C17 CR_NOISE', 'RXX_S00 C17 NOISE', 'RXX_S00 C17 PROCESSING']}, 'SUCCESS': False}
```
r-owen, line 1: Please pretty-print both of your yaml data files.

tests/rules/test_cp_verify_alarm.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def make_config(self, name):
+        """Make a config for the Cp Verify rule.
+
+        Parameters
+        ----------
+        name : `str`
+            CSC name and index in the form `name` or `name:index`.
+            The default index is 0.
+        """
```
r-owen, line 47: I'm confused by this: `name` isn't a config parameter, and all three config parameters are missing. Shouldn't the arguments be:
* calibration_type
* ocps_index
* verification_threshold

python/lsst/ts/watcher/rules/cp_verify_alarm.py
r-owen: Also
* You seem to be missing the type (integer)
* The description line above should be prefixed by `description:` or `description: >-`

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,296 @@
...
+                    description: >-
+                        Alarm calibration type: BIAS, DARK, or FLAT.
+                    type: string
+                ocps_index:
+                    OCPS index (e.g., OCPS:1: LATISS; OCPS:2: LSSTComCam).
+                verification_threshold:
+                    type: integer
+                    descriptor: Maximum number of failures per detector per test type.
+                default: 8
+            required: [calibration_type, ocps_index]
```
r-owen, line 86: Please include `verification_threshold`. It will always be present in any case, since it has a default.

r-owen (changes requested):
This is my first pass. The main thing that worries me is not running the butler query in a background thread, since the butler is not asyncio-friendly. I think your config schema may also need some tweaking.

Two more general questions:
* Do you want the operators to have to acknowledge these warnings? If not, I suggest waiting for an update Tiago requested to allow Watcher alarms to report a condition much like a log message, with no expectation that it will be acknowledged. These will be reported in a new LOVE window specifically  intended for that purpose. I haven't started work on adding the new Watcher event to ts_xml yet, but hope to get it into the next ts_xml release.
* Is Warning the only severity you want? What if there are extra failures beyond the threshold you have now? (That said, it might be simpler to get this running in its simplest form first, then add extra severities later).

python/lsst/ts/watcher/rules/cp_verify_alarm.py
plazas: Thanks! How can I specify that these values are the only ones allowed? 
We use  BIAS, DARK, or FLAT in other scripts such as https://github.com/lsst-ts/ts_externalscripts/blob/develop/python/lsst/ts/externalscripts/base_make_calibrations.py, which is the one this alarm is related to. I will leave then in uppercase if that's OK.

python/lsst/ts/watcher/rules/cp_verify_alarm.py
plazas: I added the type and the prefix. Is there an example of the use of `enum` for cases like this?

python/lsst/ts/watcher/rules/cp_verify_alarm.py
r-owen: See "enum" in jsonschema docs

python/lsst/ts/watcher/rules/cp_verify_alarm.py
r-owen: The jsonschema docs are very good

plazas: Found it, thanks!

python/lsst/ts/watcher/rules/cp_verify_alarm.py
plazas: I have added `run_in_executor` and `concurrent.futures.ThreadPoolExecutor` when getting the verification statistics, thank you for the suggestions. 
 
The idea is to run one of the instances of "base_make_calibrations.py" via LOVE every afternoon to take individual calibrations, and use these individual images to run `cp_verify` using existing standard calibrations (this is the current desired default as specified in DMTN-222, and it's called "external verification") for a given instrument. If the tests in `cp_verify` do not pass, the Watcher should tell the observer.  I'm not quite sure how to answer your questions, though.

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,305 @@
...
+        # them if verification fails.
+        thresholds = {
+            "MAX_FAILURES_PER_DETECTOR_PER_TEST_TYPE_THRESHOLD": max_number_failures_per_detector_per_test,
+            "MAX_FAILED_DETECTORS_THRESHOLD": max_number_failed_detectors,
+            "MAX_FAILED_TESTS_PER_EXPOSURE_THRESHOLD": failure_threshold_exposure,
+            "MAX_FAILED_EXPOSURES_THRESHOLD": max_number_failed_exposures,
+            "FINAL_NUMBER_OF_FAILED_EXPOSURES": failed_exposures_counter,
+        }
+
+        return certify_calib, total_counter_failed_tests, thresholds
```
r-owen, line 251: I suggest returning a dataclass with 3 named attributes. Advantages:
* Clarifies the code and makes it more robust, because you reference attributes by name, not position.
* Adds flexibility: you can add more fields later without breaking calling code.
* Simplifies early return if you specify defaults for some items -- you only need to pass the items you actually know to the constructor.

`thresholds` is another good candidate for a dataclass, depending on what you do with it. (And you can trivially turn a dataclass instance into a dict using `vars(my_dataclass_instance)`)

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,305 @@
...
+        for entry in response_verify["results"]:
+            if "verifyBiasStats" in entry["uri"]:
+                verify_stats_string = "verifyBiasStats"
+                break
+            elif "verifyDarkStats" in entry["uri"]:
+                verify_stats_string = "verifyDarkStats"
+                break
+            elif "verifyFlatStats" in entry["uri"]:
+                verify_stats_string = "verifyFlatStats"
+                break
```
r-owen, line 278: Please don't keep extracting `entry["uri"]`. Do it once at the beginning.

I also suggest a regex to simplify the code:
```
import re
...
match = re.search(r"(verify(Bias|Dark|Flat)Stats)", entry["uri"])
if match is None:
    raise ...
verify_stats_str = match.groups()[0]
```

Or if you prefer to avoid regexes, consider a loop such as:
```
uri = entry["uri"]
for substr in ("verifyBiasStats", "verifyDarkStats", "verifyFlatStats"):
    if substr in uri:
        verify_stats_string = substr
        break
else:
    raise ...
```
The `else` clause in the `for` statement is a slightly obscure Python feature. You could avoid it by initializing `verify_stats_string` to None and testing for that after the loop.

r-owen (commented):
Seems like good progress, but more work needed. I think I found at least one bug (the way you run your job as a thread). Also please look through my earlier suggestions.

tests/rules/data/cp_verify_alarm/cp_verify_bias_stats.yaml
plazas: I'd be happy to do so, but I can't find anything specific online to do this (i.e., to make a yaml file "pretty"). Did you have a package in mind or a particular standard/style that I should be following?

tests/rules/test_cp_verify_alarm.py
plazas: Ah, thanks! This probably was copied from some other alarm test many months ago. I'll change the arguments and the dictionary:  

        `config_dict = dict(calibration_type =  calibration_type,
                           ocps_index = ocps_index,
                           verification_threshold=verification_threshold
                      )`

tests/rules/data/cp_verify_alarm/cp_verify_bias_stats.yaml
plazas: Done :)

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,305 @@
...
+        # boolean: did verification fail?
+        return self.check_response(response, verify_stats)
+
+    def check_response(self, response_verify, verify_stats):
+        """Determine if cp_verify tests passed from OCPS response.
+
+        Parameters
+        ----------
+        response_verify : `dict`
+            OCPS call response.
```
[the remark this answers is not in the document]
plazas: I will add a note that points to the yaml files in `ts_watcher/tests/rules/data/cp_verify_alarm`, which are examples of these responses. Would that suffice in this case?

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,305 @@
...
+                default: 8
+            required: [calibration_type, ocps_index]
+            additionalProperties: false
+        """
+        return yaml.safe_load(schema_yaml)
+
+    def __call__(self, topic_callback):
+        msg = topic_callback.get()
+        # OCPS response after calling the pipetask
+        response = json.loads(msg.result)
```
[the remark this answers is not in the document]
plazas: Thanks! I was using `response...` to try to match the way this is generated in `https://github.com/lsst-ts/ts_externalscripts/blob/develop/python/lsst/ts/externalscripts/base_make_calibrations.py#L1549` , but I agree this is clearer.

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,305 @@
...
+        # Get the dictionary with cp_verify stats, from the butler.
+        # It might be slow, so use `run_in_executor`
+        loop = asyncio.get_running_loop()
+        verify_stats = await loop.run_in_executor(concurrent.futures.ThreadPoolExecutor(max_workers=3),
```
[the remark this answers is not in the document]
plazas: > Pass the argument(s) for get_cp_verify as additional arguments to run_in_executor.
Like this?: `verify_stats = await loop.run_in_executor(concurrent.futures.ThreadPoolExecutor(max_workers=3),
                                                  self.get_cp_verify_stats, ocps_result`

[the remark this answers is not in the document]
plazas: > If you are confident that these tasks will never build up, then I'd pass None. Otherwise I'd construct the executor once.

I have passed `None`. That function, `get_cp_verify_stats`, is parsing the OCPS result file from running `cp_verify` in order to identify 1) the calibration type and 2) call the butler with the appropriate verification collection (whose name is formed from the OCPS job ID) to get the dictionary that has the verification statistics we want to check. Unless I'm missing something, I wouldn't think this butler call will, a priori, be slow. 

`        verify_stats = await loop.run_in_executor(None,
                                                  self.get_cp_verify_stats, ocps_result)`

python/lsst/ts/watcher/rules/cp_verify_alarm.py
```diff
@@ -0,0 +1,305 @@
...
+                # the last element after splitting it.
+                fail_count = [
+                    stg.split(" ")[-1] for stg in verify_stats[exposure]["FAILURES"]
+                ]
+                counter = {}
+                for test in fail_count:
+                    if test in counter:
+                        counter[test] += 1
+                    else:
+                        counter[test] = 1
```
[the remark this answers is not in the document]
plazas: thanks!

python/lsst/ts/watcher/rules/cp_verify_alarm.py
plazas: I turned the return variables into a dataclass and I see your points. As for the `thresholds` dictionary, it's only used once to get statistics to report in the Alarm, so I'm not sure if there's much difference. I'll be happy to do it in this case too if you still think it's better.

python/lsst/ts/watcher/rules/cp_verify_alarm.py
plazas: I'm not sure if this call will be expensive. A priori, it's just one butler call per calibration type (currently `BIAS`, `DARK`, and `FLAT`) to get the verification statistics dictionary.

tests/rules/data/cp_verify_alarm/cp_verify_bias_stats.yaml
r-owen: Great! For the record, whenever I need this I search for "Yaml formatter" and pick one of the early hits.

This sort of thing is also useful for working with json schemas. Put the yaml through a yaml-to-json converter, then past that into a json schema validator (each of which is easily found with a web search). This shows any errors in the schema.

python/lsst/ts/watcher/rules/cp_verify_alarm.py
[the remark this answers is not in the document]
r-owen: It might have to, given how long and confusing those are.

python/lsst/ts/watcher/rules/cp_verify_alarm.py
r-owen: I gently suggest it, but it's OK as is.

python/lsst/ts/watcher/rules/cp_verify_alarm.py
plazas: In addition, I'm now also using `run_in_executor` for the butler creation and calls.

python/lsst/ts/watcher/rules/cp_verify_alarm.py
plazas: I turned `thresholds` into a dataclass too.

PR #58, opened 2022-12-22
Tickets/dm 37395

wvreeven:


r-owen (approved):
Thank you for making this change. (Feel free to add isort, as well, but I'm happy to do that on a later ticket).

PR #64, opened 2023-05-01
DM-38886: MTMountCsc: close and delete the model when going to fault or standby

r-owen:


bin/command_watcher
tribeiro, line 1: Do we still need these bin scripts? I though we were removing them all.

python/lsst/ts/watcher/watcher_csc.py
```diff
@@ -95,12 +96,11 @@ async def close_tasks(self):
 
     async def configure(self, config):
         if self.model is not None:
-            # this should not happen, but in case it does shut down
-            # the old model (disable alarms first because that is synchronous
-            # and may avoid old alarms appearing due to a race condition)
+            # The model should be None, but if not, close it to get rid
+            # of the old alarms.
             self.log.warning("self.model unexpectedly present in configure method")
```
tribeiro, line 101: Consider rephrasing this warning message a little bit to add more information. I'd suggest something like

```
"Model unexpectedly present in configuration stage. Closing and resetting it before continuing."
```

python/lsst/ts/watcher/watcher_csc.py
```diff
@@ -274,7 +266,18 @@ async def output_alarm(self, alarm):
         )
 
     async def handle_summary_state(self):
-        self._enable_or_disable_model()
+        if self.summary_state == salobj.State.ENABLED:
+            if self.model is None:
+                raise RuntimeError("Bug! No model.")
```
tribeiro, line 271: Considering removing the "Bug" part of the error message. Maybe something like

```
"Unable to initialize model. This is most likely due to a programatic error and might require restarting the component."
```

python/lsst/ts/watcher/watcher_csc.py
```diff
@@ -274,7 +266,18 @@ async def output_alarm(self, alarm):
...
+            self.model.enable()
+        elif self.summary_state == salobj.State.DISABLED:
+            if self.model is None:
+                raise RuntimeError("Bug! No model.")
```
tribeiro, line 275: Same as above for the error message.

python/lsst/ts/watcher/alarm.py
```diff
@@ -329,6 +332,8 @@ async def mute(self, duration, severity, user):
     async def unmute(self):
         """Unmute this alarm."""
         self._cancel_unmute()
+        if self.max_severity == AlarmSeverity.CRITICAL:
```
tribeiro, line 335: Are you sure using the `max_severity` is really what we need to do? Shouldn't it be `severity` instead? 

For example, if a CSC is in FAULT and we mute the alarm, then put it back I STANDY or DISABLED (like we have been doing with the MTMount recently) there's no need to escalate the alarm.

tests/data/config/csc/over_temperature.yaml
```diff
@@ -0,0 +1,13 @@
...
+  - name: TestOverTemperature
+    temperature_sensors:
+      - sal_index: 1
+        sensor_info:
+          - sensor_name: upstairs
+    warning_level: 19
+    # Short to speed up the unit test
+    poll_interval: 0.1
+
+
```
tribeiro, line 13: GitHub is complaining about something here, not sure what it is. Maybe it's because there are 2 empty lines?

tribeiro (commented):
This looks pretty good, I do have some comments I hope you could take a look before I approve them. 

I have a question about how you are restart the alarm after pausing..

bin/command_watcher
r-owen: That's news to me. I know they need to be trivial bin scripts that call functions to do all the work (the entry points configured in pyproject.toml) but I still use them for my own development work.

python/lsst/ts/watcher/watcher_csc.py
r-owen: I like Bug as a more efficient way to say "programming error", but I agree the message needs to be more helpful. I suggest this:

Bug: state is ENABLED but there is no model. Please restart the software and file a JIRA ticket.

tests/data/config/csc/over_temperature.yaml
r-owen: Weird. It passes pre-commit locally.

python/lsst/ts/watcher/watcher_csc.py
tribeiro: I just wonder what people will think when they see the error message. Your suggestion is fine.

python/lsst/ts/watcher/alarm.py
r-owen: We talked about this on Slack and concluded the following:
* This code has a bug: it should only restart the the escalation timer if the alarm has not been acknowledged. I will fix that.
* At present muting an alarm does not acknowledge it. That should perhaps be changed in a future ticket.
* If an alarm is displayed as unacked and has max severity CRITICAL then it should *always* be escalated (after a suitable interval, and if escalation is configured). Doing anything else would be too confusing.

PR #65, opened 2023-05-06
DM-38994: print None-state alarms when enabling the CSC

r-owen:


doc/version_history.rst
```diff
@@ -6,6 +6,25 @@
...
 ###############
 
+v1.13.2
+-------
+
+* `Model`: when enabling the model, call the alarm callback for all alarms, even those in nominal state.
+  Most alarms will usually be in nominal state.
+* `WatcherCsc`:
+
+    * When enabling the CSC: print alarm events for all alarms, even those in nominal state.
```
wvreeven, line 16: Nitpicking but in line 12 you separate the `when` part of the sentence with a `,` and in this line with a `:`. Please choose one and stick to that.

doc/version_history.rst
```diff
@@ -6,6 +6,25 @@
...
+      Most alarms will usually be in nominal state.
+    * ``showAlarms`` command: print an alarm event for all events, even those in nominal state.
+
+Requires:
+
+* ts_utils 1.1
+* ts_salobj 7.1
+* ts_idl 2
+* IDL files for ``Watcher``, ``ATDome``, ``ESS``, ``MTMount``, ``ScriptQueue``, and ``Test``, plus any additional SAL components you wish to watch.
+  These may be generated using ``make_idl_files.py`` built with ts_xml 13 and ts_sal 7.
```
wvreeven, line 26: I wonder if the `ts_xml` version you mention here is still sufficient. Did you test that? Perhaps this needs to be upped to 16 by now?

python/lsst/ts/watcher/model.py
```diff
@@ -155,23 +155,37 @@ def __init__(self, domain, config, alarm_callback=None):
...
         return self._enabled
 
-    def enable(self):
+    async def enable(self):
         """Enable the model. A no-op if already enabled."""
+        self.enable_task = asyncio.create_task(self._enable_impl())
+        await self.enable_task
+
+    async def _enable_impl(self):
+        """Implement of the enable method.
```
wvreeven, line 167: `Implement` -> `Implementation`? Or `Implement of` -> `Implement`?

wvreeven (approved):
Looks good to me. I left a few minor comments for your consideration.

doc/version_history.rst
r-owen: Good catch. I did add the `notification` event, and any rule that uses it will require ts_xml 16. We don't have any of those yet. I'll think about how to say that.

python/lsst/ts/watcher/model.py
r-owen: Ouch. Thank you for catching that!

PR #67, opened 2023-06-07
DM-37999: refine the API for BaseRule. Also, by default, reject rule topics for indexed CSCs if index=0

r-owen:


tribeiro (approved):
Looks goods..

PR #68, opened 2023-06-20
DM-39685: Add MTAirCompressor rules

r-owen:


python/lsst/ts/watcher/rules/mt_air_compressors_pressure.py
```diff
@@ -0,0 +1,162 @@
...
+                name="MTAirCompressor",
+                index=index,
+                callback_names=[],
+                poll_names=["tel_analogData"],
+            )
+            for index in SAL_INDICES
+        ]
+        super().__init__(
+            config=config,
+            name="MTAirCompressorsEnabled",
```
pkubanek, line 65: name="MTAirCompressorsPressure

python/lsst/ts/watcher/rules/mt_air_compressors_pressure.py
```diff
@@ -0,0 +1,162 @@
...
+        ]
+
+    def compute_alarm_severity(self):
+        data_list = [topic.get() for topic in self.topics]
+        pressure_index_list = [
+            (data.linePressure, data.sal_index)
+            for data in data_list
+            if data is not None
+        ]
+        if not pressure_index_list:
```
pkubanek, line 146: You construct an array which can hold two "None". not array doesn't mean it does not contain None. I believe this condition shall be if not(None in pressure_index_list):

python/lsst/ts/watcher/rules/mt_air_compressors_pressure.py
```diff
@@ -0,0 +1,162 @@
...
+            poll_duration = utils.current_tai() - self.poll_start_tai
+            if poll_duration > self.config.max_data_age:
+                return (
+                    AlarmSeverity.SERIOUS,
+                    f"No {self.topic_attr_name} data seen for {self.config.max_data_age} seconds",
+                )
+            else:
+                return watcher.NoneNoReason
+
+        pressure_index_list.sort()
```
pkubanek, line 156: I am missing timestamp comparsion. You are rerieveing data, but I don't see how you make sure those are recent data, not a history which .get() retrieve from cache.

pkubanek (changes requested):
Please at least fix alarm naming.

python/lsst/ts/watcher/rules/mt_air_compressors_pressure.py
r-owen: No, the "if data is not None" at line 144 above prevents the list from containing None.

python/lsst/ts/watcher/rules/mt_air_compressors_pressure.py
r-owen: Good catch. I'm still writing the code, but I had not yet dealt with old data.

r-owen:
This PR is not yet ready for review. Please withdraw your "needs changes" review.

python/lsst/ts/watcher/rules/mt_air_compressors_enabled.py
```diff
@@ -0,0 +1,107 @@
...
+            return
+        try:
+            self.states[data.sal_index] = salobj.State(data.summaryState)
+        except ValueError:
+            self.log.warning(
+                f"Ignoring unknown summaryState={data.summaryState} for MTAirCompressors:{data.sal_index}"
+            )
+            return
+
+        if salobj.State.ENABLED in self.states.values():
```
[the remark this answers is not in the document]
r-owen: We talked about this and have adopted the following logic:
* If both are disabled or enabled, all is well.
* If only one is disabled or enabled, warn (by default -- severity is configurable)
* If neither are disabled or enabled, go critical (by default -- severity is configurable)

The rule does not pay attention to the connectionStatus event. That's getting too complicated and it seems like we can probably do well enough without it.

python/lsst/ts/watcher/rules/mt_air_compressors_pressure.py
```diff
@@ -0,0 +1,162 @@
...
+from lsst.ts.idl.enums.Watcher import AlarmSeverity
+
+# SAL indices for MTAirCompressors to monitor.
+SAL_INDICES = (1, 2)
+
+
+class MTAirCompressorPressure(watcher.PollingRule):
+    """Check for low pressure.
+
+    This rule only reads ESS telemetry topics.
```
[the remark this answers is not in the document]
r-owen: Yes, but I dropped this rule after talking to you. We'll go with an ESS-based rule when the ESS starts reporting air pressure, and do without for now.

pkubanek (approved):
Looks good. Thanks.

PR #70, opened 2023-06-23
U/rowen/move base ess rule up to the main level

r-owen:


pkubanek (approved):
Moved BaseEssRule up, and that helps a lot - thanks.

PR #72, opened 2023-08-29
DM-40403: Update Watcher BaseESSRule to stop continuously broadcasting the same alarm.

tribeiro:


python/lsst/ts/watcher/base_ess_rule.py
```diff
@@ -231,27 +235,63 @@ def setup(self, model):
...
+        ----------
+        severity : `AlarmSeverity`
+            The alarm severity.
+        reason : `str`
+            The alarm reason
+
+        Returns
+        -------
+        `None` | `tuple`[ `AlarmSeverity`, `str` ]
+            Either `None`, if alarm is unchanged or the same values as the
```
wvreeven, line 282: `unchanged` -> `unchanged,`

wvreeven (approved):
Looks good. Maybe consider adding a test case for when the alarm doesn't change? Also see my comment for improving the readability of a sentence, which you may ignore if you prefer.

PR #75, opened 2023-11-06
DM-41544: Watcher continues to raise alarms of restarted components

tribeiro:
In addition to fixing the bug with the heartbeat alarm I also made some general improvements to the unit tests reliability with salobj-kafka.


couger01 (approved):
LGTM.

PR #79, opened 2024-08-20
Tickets/dm 44613

pothiers:
Add makeLogEntry command.

### DIFF
--- a/python/lsst/ts/watcher/alarm.py
+++ b/python/lsst/ts/watcher/alarm.py
@@ -20,13 +20,14 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 __all__ = ["Alarm"]
-
 import asyncio
 import inspect
 import logging
 
+import aiohttp
 from lsst.ts import utils
 from lsst.ts.idl.enums.Watcher import AlarmSeverity
+from lsst.ts.salobj.base import get_user_host
 
 # Default timeout for Alarm.assert_next_severity
 DEFAULT_NEXT_SEVERITY_TIMEOUT = 10
@@ -222,6 +223,53 @@ class Alarm:
         self._cancel_auto_unacknowledge()
         self._cancel_escalation_timer()
         self._cancel_unmute()
+
+    async def make_log_entry(self, log_server_url):
+        """Post message to narrative log entry in response to alarm
+        Parameters
+        ----------
+        log_server_url: `str`
+            URL of the narrativelog service.
+
+        Returns
+        -------
+        response: `dict`
+           JSON respose from Post
+
+        """
+        alarm_severity_level = {
+            AlarmSeverity.NONE: logging.DEBUG,
+            AlarmSeverity.SERIOUS: logging.INFO,
+            AlarmSeverity.WARNING: logging.WARNING,
+            AlarmSeverity.CRITICAL: logging.CRITICAL,
+        }
+
+        now = utils.astropy_time_from_tai_unix(utils.current_tai()).datetime.isoformat()
+        # Required? fields in payload:
+        #   message_text, level, user_id, user_agent, is_human
+        message = f"alarm:{self.name} severity={self.severity} {self.reason}"
+        payload = {
+            "message_text": message,
+            "level": alarm_severity_level[self.severity],
+            "user_id": get_user_host(),
+            "user_agent": "Watcher",
+            "is_human": False,
+            "tags": ["watcher", "alarm", "make_log_entry"],
+            "date_begin": now,
+            "date_end": now,
+        }
+
+        url = f"{log_server_url}/messages"
+
+        # AIOHTTP docs say don't create session per request. We do so anyhow.
+        # By not specifying a timeout, we accept the default value of
+        # 5 minutes (according to the doc)for the whole
+        # operation (connect, write, response).
+        async with aiohttp.ClientSession(raise_for_status=True) as session:
+            async with session.post(url=url, json=payload) as response:
+                response: dict = await response.json()
+                self.log.debug(f"Response (json) from Post: {response=}")
+        return response
 
     async def acknowledge(self, severity, user):
         """Acknowledge the alarm.
--- a/python/lsst/ts/watcher/config_schema.py
+++ b/python/lsst/ts/watcher/config_schema.py
@@ -26,7 +26,7 @@ import yaml
 CONFIG_SCHEMA = yaml.safe_load(
     """
 $schema: http://json-schema.org/draft-07/schema#
-title: Watcher v5
+title: Watcher v6
 description: Configuration for the Watcher
 type: object
 required:
@@ -38,6 +38,9 @@ required:
   - escalation_url
 additionalProperties: false
 properties:
+  narrative_server_url:
+    description: URL of narrativelog service
+    type: string
   disabled_sal_components:
     description: >-
       Names of SAL components that the Watcher will ignored.
--- a/python/lsst/ts/watcher/model.py
+++ b/python/lsst/ts/watcher/model.py
@@ -216,6 +216,19 @@ class Model:
         self.disable()
         await asyncio.gather(*[remote.close() for remote in self.remotes.values()])
 
+    async def make_log_entry(self, name):
+        """MakeLogEntry for alarm.
+
+        Parameters
+        ----------
+        name : `str`
+            Regular expression for alarm name(s) to post to narrative log.
+        """
+        log_server = self.config.narrative_server_url
+
+        for rule in self.get_rules(name):
+            await rule.alarm.make_log_entry(log_server)
+
     async def acknowledge_alarm(self, name, severity, user):
         """Acknowledge one or more alarms.
 
--- a/python/lsst/ts/watcher/watcher_csc.py
+++ b/python/lsst/ts/watcher/watcher_csc.py
@@ -28,7 +28,6 @@ from http import HTTPStatus
 
 import aiohttp
 from lsst.ts import salobj
-from lsst.ts.xml.component_info import ComponentInfo
 
 from . import __version__
 from .config_schema import CONFIG_SCHEMA
@@ -73,16 +72,6 @@ class WatcherCsc(salobj.ConfigurableCsc):
         # and reset to None when the Watcher goes to standby.
         self.model = None
         self.http_client = aiohttp.ClientSession()
-
-        # TODO DM-44613: Remove compatibility commands
-        compatibility_commands = dict(
-            makeLogEntry=self._do_makeLogEntry,
-        )
-
-        component_info = ComponentInfo("Watcher", "none")
-        for command_name, command_method in compatibility_commands.items():
-            if f"cmd_{command_name}" in component_info.topics:
-                setattr(self, f"do_{command_name}", command_method)
 
         super().__init__(
             "Watcher",
@@ -336,12 +325,10 @@ class WatcherCsc(salobj.ConfigurableCsc):
         self.assert_enabled()
         await self.model.unmute_alarm(name=data.name)
 
-    async def _do_makeLogEntry(self, data):
+    async def do_makeLogEntry(self, data):
         """Make log entry for alarms."""
         self.assert_enabled()
-        # TODO DM-44613: Implement watcher command to create
-        # log entry for a particular alarm.
-        raise NotImplementedError("Command not implemented yet.")
+        await self.model.make_log_entry(name=data.name)
 
 
 def run_watcher():
--- a/tests/data/config/csc/_init.yaml
+++ b/tests/data/config/csc/_init.yaml
@@ -5,3 +5,4 @@ rules: []
 escalation: []
 escalation_url: ""
 escalation_timeout: 1
+narrative_server_url: https://fake.url.com/narrativelog
--- a/tests/test_csc.py
+++ b/tests/test_csc.py
@@ -24,6 +24,7 @@ import os
 import pathlib
 import sys
 import unittest
+from unittest.mock import AsyncMock, patch
 
 import pytest
 from lsst.ts import salobj, utils, watcher
@@ -151,7 +152,7 @@ class CscTestCase(salobj.BaseCscTestCase, unittest.IsolatedAsyncioTestCase):
             desired_config_pkg_name = "ts_config_ocs"
             desired_config_env_name = desired_config_pkg_name.upper() + "_DIR"
             desird_config_pkg_dir = os.environ[desired_config_env_name]
-            desired_config_dir = pathlib.Path(desird_config_pkg_dir) / "Watcher/v5"
+            desired_config_dir = pathlib.Path(desird_config_pkg_dir) / "Watcher/v6"
             assert self.csc.get_config_pkg() == desired_config_pkg_name
             assert self.csc.config_dir == desired_config_dir
 
@@ -187,6 +188,7 @@ class CscTestCase(salobj.BaseCscTestCase, unittest.IsolatedAsyncioTestCase):
                     "showAlarms",
                     "unacknowledge",
                     "unmute",
+                    "makeLogEntry",
                 ),
                 override="two_scriptqueue_enabled.yaml",
             )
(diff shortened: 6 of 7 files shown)


python/lsst/ts/watcher/alarm.py
```diff
@@ -223,6 +230,42 @@ def close(self):
...
+        payload = {
+            "message_text": msg,
+            "level": 0,
+            "user_id": "admin@love1.tu.lsst.org",  # TODO
+            "user_agent": user,
+            "is_human": False,
+            "tags": ["watcher", "alarm", "make_log_entry"],
+            "date_begin": now,
+            "date_end": now,
+            # ## Let following fields be defaulted.
```
tribeiro, line 249: Please, remove commented lines.

python/lsst/ts/watcher/alarm.py
```diff
@@ -223,6 +230,42 @@ def close(self):
...
+            # 'systems': ['ObservatorySoftwareTools'],
+            # 'subsystems': ['LOVE'],  # TODO
+            # 'cscs': [],
+            # 'urls': [],
+            # 'time_lost': 0,
+        }
+
+        url = f"{log_server_url}/messages"
+        print(f"Posting to log: {url=} {payload=}")  # TODO REMOVE!!!
```
tribeiro, line 258: Remove this line or use `self.log.debug`.

python/lsst/ts/watcher/alarm.py
```diff
@@ -223,6 +230,42 @@ def close(self):
...
+
+        # AIOHTTP docs say don't create session per request. We do so anyhow.
+        # Use default timeout (=5 minutes for whole operation)
```
tribeiro, line 261: I don't see any timeout below. Also, 5 minutes is a lot! I would expect at most 10-15s.

python/lsst/ts/watcher/alarm.py
```diff
@@ -223,6 +230,42 @@ def close(self):
...
+        async with aiohttp.ClientSession(raise_for_status=True) as session:
+            async with session.post(url=url, json=payload) as response:
+                post_values: dict = await response.json()
+                dbg_msg = f"Response (json) from Post: {post_values=}"
+                self.log.debug(dbg_msg)
```
tribeiro, line 266: Any reason for not doing

```
self.log.debug(f"Response (json) from Post: {post_values=}.")
```

?

python/lsst/ts/watcher/alarm.py
```diff
@@ -223,6 +230,42 @@ def close(self):
         self._cancel_escalation_timer()
         self._cancel_unmute()
 
+    async def make_log_entry(self, log_server_url):
+        """Post message to narrative log entry in response to alarm"""
```
tribeiro, line 234: Please document the input and return values.

python/lsst/ts/watcher/config_schema.py
```diff
@@ -38,6 +38,10 @@
   - escalation_url
 additionalProperties: false
 properties:
+  narrative_server_url:
+    description: >-
```
tribeiro, line 42: You can make this in one line:

```
description: Url of narrativelog service.
```

python/lsst/ts/watcher/model.py
```diff
@@ -216,6 +216,21 @@ async def close(self):
...
 
+    async def make_log_entry(self, name):
+        """MakeLogEntry for alarm.
+
+        Parameters
+        ----------
+        name : `str`
+            Regular expression for alarm name(s) to post to narrative log.
+        """
+        print(f"model.make_log_entry: {name=} {self.config=}")
```
tribeiro, line 227: Please, either use `self.log.debug` or remove the print statement.

python/lsst/ts/watcher/model.py
```diff
@@ -216,6 +216,21 @@ async def close(self):
...
+        log_server = self.config.narrative_server_url
+        print(f"model.make_log_entry: {log_server=}")
```
tribeiro, line 229: Same as above either use `self.log.debug` or remove print.

python/lsst/ts/watcher/watcher_csc.py
```diff
@@ -28,12 +28,14 @@
 
 import aiohttp
 from lsst.ts import salobj
-from lsst.ts.xml.component_info import ComponentInfo
 
 from . import __version__
 from .config_schema import CONFIG_SCHEMA
 from .model import Model
 
+# from lsst.ts.xml.component_info import ComponentInfo
```
tribeiro, line 36: Remove commented line

tests/data/config/csc/_init.yaml
```diff
@@ -5,3 +5,4 @@ rules: []
 escalation: []
 escalation_url: ""
 escalation_timeout: 1
+narrative_server_url: https://tucson-teststand.lsst.codes/narrativelog
```
tribeiro, line 8: Maybe use a fake url? Just to make sure people don't think we are hitting the TTS backend for testing.

tests/test_csc.py
```diff
@@ -815,6 +816,25 @@ async def test_show_alarms(self):
                 assert data.acknowledgedBy == ""
             assert seen_alarm_names == alarm_names
 
+    @patch("aiohttp.ClientSession.post")
+    async def test_make_log_entry(self, mock_post):
+        """Test the makeLogEntry command."""
+        test_case_json = {"key": "TEST-123", "status": "PASS"}
+        # mock_response = MagicMock()
```
tribeiro, line 824: Please, remove commented code.

tests/test_csc.py
```diff
@@ -815,6 +816,25 @@ async def test_show_alarms(self):
...
+
+        # Send the makeLogEntry command.
+        async with self.make_csc(
+            config_dir=TEST_CONFIG_DIR, initial_state=salobj.State.STANDBY
+        ):
+            await salobj.set_summary_state(
+                self.remote, state=salobj.State.ENABLED, override="enabled.yaml"
+            )
+            await self.check_all_alarms_events_are_none()
+            await self.remote.cmd_makeLogEntry.start(timeout=STD_TIMEOUT)
```
tribeiro, line 837: I think you want to specify an alarm to log and then check that the mock_reponse was called with the values you expected?

python/lsst/ts/watcher/alarm.py
```diff
@@ -223,6 +230,42 @@ def close(self):
...
+        now = datetime.datetime.now().isoformat()  # use ts_utils repo?
+        # Required? fields in payload:
+        #   message_text, level, user_id, user_agent, is_human
+        msg = f"alarm:{self.name} severity={self.severity}"
+        user = "make_log_entry"
```
tribeiro, line 239: User should probably be "Watcher".

python/lsst/ts/watcher/alarm.py
```diff
@@ -223,6 +230,42 @@ def close(self):
...
+        payload = {
+            "message_text": msg,
+            "level": 0,
+            "user_id": "admin@love1.tu.lsst.org",  # TODO
```
tribeiro, line 243: you can use something like this:

```
from lsst.ts.salobj.base import get_user_host

...
            "user_id": get_user_host(),
```

python/lsst/ts/watcher/alarm.py
```diff
@@ -223,6 +230,42 @@ def close(self):
...
+    async def make_log_entry(self, log_server_url):
+        """Post message to narrative log entry in response to alarm"""
+        now = datetime.datetime.now().isoformat()  # use ts_utils repo?
+        # Required? fields in payload:
+        #   message_text, level, user_id, user_agent, is_human
+        msg = f"alarm:{self.name} severity={self.severity}"
+        user = "make_log_entry"
+        payload = {
+            "message_text": msg,
+            "level": 0,
```
tribeiro, line 242: Consider doing something like:

```
alarm_severity_level = {
    AlarmSeverity.NONE: logging.DEBUG,
    AlarmSeverity.SERIOUS: logging.INFO,
    AlarmSeverity.WARNING: logging.WARNING,
    AlarmSeverity.CRITICAL: logging.CRITICAL,
}

"level": alarm_severity_level[self.severity]
```

python/lsst/ts/watcher/alarm.py
```diff
@@ -223,6 +230,42 @@ def close(self):
         self._cancel_escalation_timer()
         self._cancel_unmute()
 
+    async def make_log_entry(self, log_server_url):
+        """Post message to narrative log entry in response to alarm"""
+        now = datetime.datetime.now().isoformat()  # use ts_utils repo?
```
tribeiro, line 235: Instead of using `date time` , use `utils.current_tai()`.

You can get it iso-formatted with:

```
now = utils.astropy_time_from_tai_unix(utils.current_tai()).isoformat()
```

python/lsst/ts/watcher/alarm.py
```diff
@@ -223,6 +230,42 @@ def close(self):
...
+        # Required? fields in payload:
+        #   message_text, level, user_id, user_agent, is_human
+        msg = f"alarm:{self.name} severity={self.severity}"
```
tribeiro, line 238: Please add the `self.reason` to the message e.g.;

```
msg = f"alarm:{self.name} severity={self.severity.name} {self.reason}"
```


Also, expand it from `msg` to `message`, maybe even `message_text`, such that you would have:

```
message_text: message_text,
```

tribeiro (changes requested):
Here some general comments.

python/lsst/ts/watcher/alarm.py
pothiers: The comment means that by not specifying a timeout, I accepted the default which is 5 minutes according to the doc.  This is for whole operation (connect, write, response).

python/lsst/ts/watcher/alarm.py
pothiers: Changed.  But only reason was I was using dbg_msg in both log and print.  I don't know where the log is, so print was easier for me during dev.

python/lsst/ts/watcher/alarm.py
pothiers: Changed.  But is there a reason for this requested change?  The original does it more simply.  Using the standard python library makes it easier for someone unfamiliar to all the ts code to know what it does.

tests/test_csc.py
pothiers: How do I specify an alarm to log?

python/lsst/ts/watcher/alarm.py
pothiers: I tried your alternative.  But it got an error:
> AttributeError: 'Time' object has no attribute 'isoformat'

I reverted back to use datetime.

pothiers:
I've pushed changes to everything except 
https://github.com/lsst-ts/ts_watcher/pull/79#discussion_r1724847918
I'm unfamiliar with how to use the alarm system in general and within a test in particular.  If you can tell me how to specify an alarm I'll add change as well.

python/lsst/ts/watcher/alarm.py
tribeiro: Sorry, I missed one thing on my suggested code, it should be:

```
now = utils.astropy_time_from_tai_unix(utils.current_tai()).datetime.isoformat()
```

There are a couple reasons I think this is a better alternative;

1. In general we use the lsst.ts utility to manage timestamps.
2. It will make sure the timestamp has the correct format (TAI).
3. You don't need to import any additional library.

python/lsst/ts/watcher/alarm.py
tribeiro: Maybe expand your comment to clarify this point then? it is not clear from the text above. I would suggest something like:

```
        # AIOHTTP docs say don't create session per request. We do so anyhow.
        # By not specifying a timeout, we accept the default value of 5 minutes (according to the doc)
        # for the whole operation (connect, write, response).
```

tests/test_csc.py
tribeiro: It would be something like this:

```
await self.remote.cmd_makeLogEntry.set_start(
    name="Enabled. ATDome:0",
    timeout=STD_TIMEOUT
)
```

To expand on the values above.

The test enables the CSC with the `enabled.yaml` configuration, which you can find [here](https://github.com/lsst-ts/ts_watcher/blob/develop/tests/data/config/csc/enabled.yaml).

In there you can see that is configures the "Enabled" [rule](https://github.com/lsst-ts/ts_watcher/blob/eca6e70b842d6e651e2f19edf58e21132f5a4c66/tests/data/config/csc/enabled.yaml#L6) for 3 CSCs, ATCamera, ATDome and ScriptQueue:2. However it disables the rule for ATCamera.

By passing `Enabled. ATDome:0` as the `name` of the alarm to make a log entry for, it means it will log the entry for that alarm in particular. You should be able to force that alarm to be critical by doing this:

```
# Send the makeLogEntry command.
async with self.make_csc(
    config_dir=TEST_CONFIG_DIR, initial_state=salobj.State.STANDBY
), salobj.Controller(name="ATDome", write_only=True) as atdome:

    atdome_alarm_name = "Enabled.ATDome:0"
    await self.check_all_alarms_events_are_none()

    await atdome.evt_summaryState.set_write(
        summaryState=salobj.State.FAULT, force_output=True
    )
    await self.assert_next_alarm(
        name=atdome_alarm_name,
        severity=AlarmSeverity.CRITICAL,
        maxSeverity=AlarmSeverity.CRITICAL,
        acknowledged=False,
        acknowledgedBy="",
    )

    await self.remote.cmd_makeLogEntry.set_start(
        name=atdome_alarm_name,
        timeout=STD_TIMEOUT,
    )

    # assert here that mock_response got the expected payload.

```

tests/test_csc.py
```diff
@@ -815,6 +817,91 @@ async def test_show_alarms(self):
                 assert data.acknowledgedBy == ""
             assert seen_alarm_names == alarm_names
 
+    @patch("aiohttp.ClientSession.post")
+    async def test_make_log_entry(self, mock_post):
+        """Test the makeLogEntry command."""
+
+        def dict_subset(smallDict, largeDict):
```
tribeiro, line 824: method parameters should be `snake_case`...

However, do you really need a method just for that? seems like unnecessary overhead for something you can do inline.

tests/test_csc.py
```diff
@@ -815,6 +817,91 @@ async def test_show_alarms(self):
...
+            test_response = await self.remote.cmd_makeLogEntry.set_start(
+                name=atdome_alarm_name,
+                timeout=STD_TIMEOUT,
+            )
+
+            # The actual test_response will have more fields than expected.
+            # We don't care about most of them and they might cause variations
+            # in results so construct a subset of the the actual for only the
+            # keys in expected.
+            expected_response = {
```
tribeiro, line 864: You can remove this test here of the `test_reponse`.

tests/test_csc.py
```diff
@@ -815,6 +817,91 @@ async def test_show_alarms(self):
...
+            expected = {
+                "url": "https://fake.url.com/narrativelog/messages",
+                "json": {
+                    "message_text": "alarm:Enabled.ATDome:0 severity=4 FAULT state",
+                    "level": 50,
+                    "user_id": "saluser@chimp20",
+                    "user_agent": "Watcher",
+                    "is_human": False,
+                    "tags": ["watcher", "alarm", "make_log_entry"],
+                    "date_begin": "2024-08-22T15:22:37.444279",
```
tribeiro, line 890: I think you can remove this one from the expected.

tests/test_csc.py
```diff
@@ -815,6 +817,91 @@ async def test_show_alarms(self):
...
+                    "date_end": "2024-08-22T15:22:37.444279",
```
tribeiro, line 891: I think you can remove this one from the expected.

tests/test_csc.py
```diff
@@ -815,6 +817,91 @@ async def test_show_alarms(self):
...
+                },
+            }
+            self.assertIn("user_id", post_args["json"])
+            self.assertIn("date_begin", post_args["json"])
+            self.assertIn("date_end", post_args["json"])
+            del post_args["json"]["user_id"]
+            del post_args["json"]["date_begin"]
+            del post_args["json"]["date_end"]
+            del expected["json"]["user_id"]
```
tribeiro, line 900: Instead of adding and then deleting, just don't add these to the expected dict.

tribeiro (commented):
Alright! it is looking really good! I only have a few more comments and then I can approved the PR.

tests/test_csc.py
pothiers: I would prefer to have the full data  that is returned in expected. It IS what is expected, after all.  Its just that we are less picky about a few of the expected values.  Think keeping Expected and Actual (post_args) in sync makes things more obvious and more maintainable.

tests/test_csc.py
pothiers: I think the overhead is insignificant compared to issues of understandability, readability, and reuse. I've used this technique before and expect I will again.  Maybe someone will use it in this file.

tests/test_csc.py
pothiers: Why? If my cmd_makeLogEntry was broken in a certain way, wouldn't that response be different?

tribeiro: no.. if your command was broken it would raise an exception

tests/test_csc.py
tribeiro: Well, but the problem is you "expected" is not really what you expected right? The dates are dynamic so your expectation is that the fields will exists, but you can't determine the values.

Think about someone looking at this in the future. They will wonder, why there's this hard-coded date here in the "expected" dictionary and why it is being deleted afterwards?

tests/test_csc.py
pothiers: I guess you are saying that there is no possibility of it returning a different value of any keys in the response.  

Maybe we would want underlying command to, for example, notice that the narrativelog API returned something different than the posted payload (its supposed to return the full record that was inserted) and indicate so with a different response.

tests/test_csc.py
pothiers: Yes, they might wonder.  I would actually prefer to comment about the part of the structure in expected that we are not using.  But I thought you would object to such commented code.  Personally, if I saw a hard-coded date, I would think:  "Oh, of course, it varies so cannot compare to it.  But I'm glad they showed me an example of it so I know exactly the format."

tests/test_csc.py
tribeiro: >I guess you are saying that there is no possibility of it returning a different value of any keys in the response.

yes

tests/test_csc.py
tribeiro: Commenting code is against the guidelines so, yeah, probably not a good idea. However, you could make a comment just above the `expected` dictionary saying that the dictionary also contains the additional keys that are dynamic, and therefore, removed from the expected dictionary.

tribeiro (approved):
Looks good! thanks for all the changes!

PR #80, opened 2024-08-21
Add `rules.PowerOutage` rule.

wvreeven:


python/lsst/ts/watcher/rules/power_outage.py
```diff
@@ -0,0 +1,124 @@
...
+  - min_num_zeros_schneider
+additionalProperties: false
+       """
+        return yaml.safe_load(schema_yaml)
+
+    def compute_alarm_severity(
+        self, data: salobj.BaseMsgType, **kwargs: typing.Any
+    ) -> AlarmSeverityReasonType:
+        severity, reason = NoneNoReason
+        if hasattr(data, "activePowerA"):
```
tribeiro, line 99: So, these will most likely happen at the same time right? And they will also be a continuous occurrence, based on the data you showed in the ticket. I would suggest you consider we configure the rule to look for one or the other. Otherwise this alarm will be constantly activating jumping from one set of UPSs to the other.

tribeiro (commented):
I have one comment about how you setup the alarm. I would suggest this could be configured as either one or another or we will be constantly flooded by the alarm being activated.

tribeiro (approved):
Alright, this looks good.. can you please add docstrings to the methods you introduced?

PR #81, opened 2024-08-28
Tickets/dm 45056

wvreeven:


python/lsst/ts/watcher/rules/mtdome_az_enabled.py
```diff
@@ -0,0 +1,106 @@
...
+            `update_alarm_severity`, the arguments will be as follows:
+
+            * topic_callback : `TopicCallback`
+              Topic callback wrapper.
+
+        Returns
+        -------
+        None, if no change or unknown, or a tuple of two values:
+
+        severity: `lsst.ts.idl.enums.Watcher.AlarmSeverity`
```
tribeiro, line 87: These are no longer from idl but from xml.

python/lsst/ts/watcher/rules/mtdome_az_enabled.py
```diff
@@ -0,0 +1,106 @@
...
+            to ``nominal`` state.
+
+        Notes
+        -----
+        You may return `NoneNoReason` if the alarm state is ``NONE``.
+        """
+        severity, reason = NoneNoReason
+
+        if data.faultCode != "":
+            reason = data.faultCode
```
tribeiro, line 103: There's a `status` attribute in this topic, which I think is an enumeration. I would suggest you add this information to the reason string, just to add more information. Maybe something like:

```
reason = f"Dome motion state {MotionState(data.state)!r}: {data.faultCode}."
```

python/lsst/ts/watcher/rules/mtdome_az_enabled.py
```diff
@@ -0,0 +1,106 @@
...
+            what value is out of range, and what the range is.
+            If ``severity`` is ``NONE`` then this value is ignored (but still
+            required) and the old reason is retained until the alarm is reset
+            to ``nominal`` state.
+
+        Notes
+        -----
+        You may return `NoneNoReason` if the alarm state is ``NONE``.
+        """
+        severity, reason = NoneNoReason
```
tribeiro, line 100: Minor picky comment but I think you can accomplish all this in a one liner:

```
return NoneNoReason if data.faultCode else AlarmSeverity.CRITICAL, f"Dome motion state {MotionState(data.state)!r}: {data.faultCode}."
```

python/lsst/ts/watcher/rules/mtdome_capacitor_banks.py
```diff
@@ -0,0 +1,119 @@
...
+            RemoteInfo(
+                name=remote_name,
+                index=remote_index,
+                callback_names=["evt_capacitorBanks"],
+                poll_names=[],
+            )
+        ]
+        super().__init__(
+            config=config,
+            name=f"MTDomeCapacitorBanks.{remote_name}:{remote_index}",
```
tribeiro, line 65: The MTDome is not indexed, I would suggest dropping the index from the name.

python/lsst/ts/watcher/rules/mtdome_az_enabled.py
```diff
@@ -0,0 +1,106 @@
...
+            RemoteInfo(
+                name=remote_name,
+                index=remote_index,
+                callback_names=["evt_azEnabled"],
+                poll_names=[],
+            )
+        ]
+        super().__init__(
+            config=config,
+            name=f"MTDomeAzEnabled.{remote_name}:{remote_index}",
```
tribeiro, line 57: The MTDome is not indexed, I suggest dropping the index from the name.

python/lsst/ts/watcher/rules/mtdome_capacitor_banks.py
```diff
@@ -0,0 +1,119 @@
...
+        You may return `NoneNoReason` if the alarm state is ``NONE``.
+        """
+        severity, reason = NoneNoReason
+
+        reason_list: list[str] = []
+        for attribute, value in vars(data).items():
+            # For any of the attributes, value is a list of bool.
+            if attribute in REASON_DICT and True in value:
+                reason_list.append(f"{REASON_DICT[attribute]} detected")
+        if len(reason_list) > 0:
```
tribeiro, line 115: I think you can do `if reason_list:` only.. The DM style guide suggests we use this style for [sequence comparison](https://developer.lsst.io/python/style.html#comparisons).

python/lsst/ts/watcher/rules/mtdome_capacitor_banks.py
```diff
@@ -0,0 +1,119 @@
...
+            to ``nominal`` state.
+
+        Notes
+        -----
+        You may return `NoneNoReason` if the alarm state is ``NONE``.
+        """
+        severity, reason = NoneNoReason
+
+        reason_list: list[str] = []
+        for attribute, value in vars(data).items():
```
tribeiro, line 111: I think it would be simpler to just loop over the `REASON_DICT ` dictionary and use `getattr` no? Also, the way it is done here, there's a chance that the xml is updated to rename a field and we will never get a failure to indicate the change.

tribeiro (approved):
looks good, I left a few inline comments I hope you will consider before merging. They are mostly nitty picky details so feel free to ignore what you don't think is relevant.

python/lsst/ts/watcher/rules/mtdome_az_enabled.py
wvreeven: I think it should be `return NoneNoReason if not data.faultCode else ...`. Great suggestion!

PR #82, opened 2024-08-29
Add the mt_vibration.py.

teweitsai:
* Add the `rules.MTVibration` rule.

### DIFF
--- a/doc/version_history.rst
+++ b/doc/version_history.rst
@@ -6,6 +6,11 @@
 Version History
 ###############
 
+v1.18.1
+-------
+
+* Add the `rules.MTVibrationRotator` rule.
+
 v1.18.0
 -------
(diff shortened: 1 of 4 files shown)


b1quint:
@teweitsai I think we will have more alarms associated with vibrations on the TMA. I wonder if we want a more specific name for the class to allow the implementation of similar alarms on other components like hexapods, VMS, etc. What do you think?

teweitsai:
I renamed the rule to be more specific in the commit: [5f5fc69](https://github.com/lsst-ts/ts_watcher/pull/82/commits/5f5fc69cc97a0bd70d0a7289152c88527cb549fc). Thanks!

tribeiro (approved):
looks good.

PR #84, opened 2024-09-24
Improve the MTTangentLinkTemperature and MTForceError.

teweitsai:


python/lsst/ts/watcher/rules/mt_tangent_link_temperature.py
```diff
@@ -169,19 +169,22 @@ def compute_alarm_severity(self) -> AlarmSeverityReasonType:
         temperature_tangent = temperature_tangent_contain_nan[
             ~np.isnan(temperature_tangent_contain_nan)
         ]
+        temperature_tangent.sort()
 
         temperature_ring = np.array(self._remote_m2.tel_temperature.get().ring)
 
+        # Only consider the 3 active tangent links and ignore the other 3
```
tribeiro, line 176: Isn't it still worth to monitor the temperature of all tangent links?

tribeiro (approved):
Looks good. I think however you should still monitor all tangent links instead of limiting it to the active ones. We don't really gain anything in excluding the links but potentially get the possibility of finding issues with the other links.

python/lsst/ts/watcher/rules/mt_tangent_link_temperature.py
teweitsai: Improved in the commit: [fdafebf](https://github.com/lsst-ts/ts_watcher/pull/84/commits/fdafebf3ccfb878d63868c4dd6a073433a99323d). Thanks!

PR #86, opened 2024-10-30
Add the `rules.MTMirrorSafety` rule.

wvreeven:


python/lsst/ts/watcher/threshold_handler.py
```diff
@@ -213,17 +233,53 @@ def get_severity_reason(self, value, current_severity, source_descr):
...
+                    return make_severity_reason(
+                        severity=severity, with_hysteresis=False
+                    )
 
         return NoneNoReason
 
+    def get_current_tai(self) -> float:
+        """Get the current TAI time [UNIX seconds].
+
+        This method is designed to be overridden in unit tests.
```
tribeiro, line 273: you could use pytest.monkeypatch instead.

tribeiro (approved):
Looks goos thanks!

python/lsst/ts/watcher/threshold_handler.py
wvreeven: Thanks for the tip. I'll keep that in mind for a future implementation.

PR #88, opened 2024-11-13
Add the `rules.Telemetry` rule.

wvreeven:


python/lsst/ts/watcher/rules/telemetry.py
```diff
@@ -0,0 +1,174 @@
...
+        self, data: salobj.BaseMsgType, **kwargs: typing.Any
+    ) -> AlarmSeverityReasonType:
+        topic_callback = kwargs["topic_callback"]
+        _, _, topic_name = topic_callback.topic_key
+        if topic_name == "evt_summaryState":
+            self.log.debug(
+                f"Received evt_summaryState for {self.name} with summaryState={data.summaryState}"
+            )
+            self.csc_should_receive_telemetry = data.summaryState in [
+                salobj.State.DISABLED,
```
tribeiro, line 136: Note that not all CSCs publish telemetry when they are in DISABLED. You might want to make this configurable.

tribeiro (approved):
Looks good, but please, see my last comment.

PR #91, opened 2025-01-15
Tickets/dm 46560

teweitsai:
* Add the `rules.MTHexapodHighCurrent` and `rules.MTHexapodTransitionToIdle` rules.

### DIFF
--- a/python/lsst/ts/watcher/rules/mt_hexapod_high_current.py
+++ b/python/lsst/ts/watcher/rules/mt_hexapod_high_current.py
@@ -1,0 +1,121 @@
+# This file is part of ts_watcher.
+#
+# Developed for Vera C. Rubin Observatory Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = ["MTHexapodHighCurrent"]
+
+
+import logging
+import types
+import typing
+
+from lsst.ts import salobj, watcher
+from lsst.ts.xml.enums.MTHexapod import SalIndex
+from lsst.ts.xml.enums.Watcher import AlarmSeverity
+
+from ..base_rule import AlarmSeverityReasonType, NoneNoReason
+from ..remote_info import RemoteInfo
+
+
+class MTHexapodHighCurrent(watcher.BaseRule):
+    """Monitor the high motor current event of the MT hexapod.
+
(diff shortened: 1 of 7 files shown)


python/lsst/ts/watcher/rules/mt_hexapod_transition_to_idle.py
tribeiro, line 1: Oh, I don't know if we want an alarm for this. Why would we need to alert the OSs that this is happening? Not to mention that this is probably going to happen pretty frequently.

python/lsst/ts/watcher/rules/mt_hexapod_high_current.py
```diff
@@ -0,0 +1,121 @@
...
+    log : `logging.Logger` or None, optional
+        Parent logger. (the default is None)
+    """
+
+    def __init__(
+        self, config: types.SimpleNamespace, log: logging.Logger | None = None
+    ):
+
+        remote_name = "MTHexapod"
+        remote_info_camera = RemoteInfo(
```
tribeiro, line 53: I am not sure compounding the 2 hexapods into the same rule is a good idea. It means we cannot independently control muting/acknowledging etc of the alarms.  I think we probably want an alarm that takes the CSC index (or name->index mapping) as a configuration.

python/lsst/ts/watcher/rules/mt_hexapod_high_current.py
```diff
@@ -0,0 +1,121 @@
...
+            If ``severity`` is ``NONE`` then this value is ignored (but still
+            required) and the old reason is retained until the alarm is reset
+            to ``nominal`` state.
+
+        Notes
+        -----
+        You may return `NoneNoReason` if the alarm state is ``NONE``.
+        """
+
+        sal_index = SalIndex(data.salIndex)
```
tribeiro, line 110: yeah, I am not really sure I agree with this implementation. Why not let the alarm read the current and compute the alarm severity here? I don't think relying on the CSC to send this is a good idea.

tribeiro (commented):
So, I think we need to take a step back on this one here. 

For the high current alarm, I really think we could have the alarm calculate the alarm level itself instead of relying on a flag from the CSC. 

For the transition to idle one, I really don't think this is something we should be bothering the OSs with.

python/lsst/ts/watcher/rules/mt_hexapod_transition_to_idle.py
teweitsai: The main reason is that Doug suggested that since the idle time is not too short, people should rewarm the hexapod after the idle before the reuse. BTW, It is not healthy to have a short idle time for the hardware. Therefore, my original plan is to have an alarm to remind the users that they need to re-execute the warm up script. This logic is easier to perform in the hexapod CSC level. But we could try to put the complexity in the watcher level as well if you want. Thanks!

python/lsst/ts/watcher/rules/mt_hexapod_high_current.py
teweitsai: Even in the normal using, the hexapod can have the current values to be higher than the threshold. This is normal because the Copley controllers needs to have a significant torque to overcome the high friction as a reality of design flaw we have. My original plan was to use the hexapod CSC to work on this logic because it would be easier. But we could try to put the complexity in the watcher as well although it would need to base on multiple events and telemetry to judge this. Thanks!

teweitsai:
As the suggestion, I have closed the PR of `ts_xml` and remove all the logic in the hexapod CSC to judge the events. Instead, we will try to implement the related logic in the watcher level only based on the multiple events and telemetry to check the hexapod's condition. This would make the logic complex but might be doable. Therefore, I will close this PR and use another ticket to implement the alarms instead. Thanks!

PR #92, opened 2025-01-29
DM-48621: Fix issue with ATCameraDewar Watcher alarm that prevented it from resetting after a data stream interruption.

tribeiro:


pyproject.toml
```diff
@@ -33,3 +33,43 @@ asyncio_mode = "auto"
 
 [project.optional-dependencies]
 dev = ["documenteer[pipelines]"]
+
+[tool.towncrier]
```
wvreeven, line 37: There is a config in ts_pre_commit_conf as well. Please use that instead. If you disagree, please explain why.

wvreeven (approved):
LGTM. Please let's make a decision on how and where to store the towncrier configuration.

PR #94, opened 2025-02-13
Add the rule of hexapod overcurrent event.

teweitsai:


python/lsst/ts/watcher/base_hexapod_overcurrent_rule.py
```diff
@@ -0,0 +1,190 @@
...
+        Parent logger. (the default is None)
+    """
+
+    def __init__(
+        self,
+        salIndex: SalIndex,
+        config: types.SimpleNamespace,
+        log: logging.Logger | None = None,
+    ):
+        is_camera_hexapod = salIndex == SalIndex.CAMERA_HEXAPOD
```
tribeiro, line 56: Please, consider making the index or, even better, the "name" of the hexapod (as in the `SalIndex` enumeration) a configuration parameter. Then you won't need to subclass it.

python/lsst/ts/watcher/base_hexapod_overcurrent_rule.py
```diff
@@ -0,0 +1,190 @@
...
+            type: object
+            properties:
+                threshold_current:
+                    description: >-
+                        Threshold of the current in ampere.
+                    type: number
+                    default: 4.0
+                max_count:
+                    description: >-
+                        Maximum count to check the hexapod current in idle.
```
tribeiro, line 96: wouldn't it be easier to make this the "time window" (e.g. 10 minutes) instead?

python/lsst/ts/watcher/base_hexapod_overcurrent_rule.py
```diff
@@ -0,0 +1,190 @@
...
+            self._count += 1
+        else:
+            self._count = 0
+
+        return (
+            NoneNoReason
+            if (self._count < self.config.max_count)
+            else (
+                AlarmSeverity(self.config.severity),
+                f"{self._hexapod} has the overcurrent.",
```
tribeiro, line 175: Consider something like:

```
f"{self._hexapod} overcurrent above threshold.
```

or something like that...

python/lsst/ts/watcher/base_hexapod_overcurrent_rule.py
```diff
@@ -0,0 +1,190 @@
...
+        # In the normal operation, the motor current can be higher than the
+        # threshold. Therefore, we only care about the condition that we always
+        # have the motor current to be higher than the threshold all the time
+        # when there is no movement.
+        if self._is_enabled_and_stationary() and any(
+            [current >= self.config.threshold_current for current in data.motorCurrent]
+        ):
+            self._count += 1
+        else:
+            self._count = 0
```
tribeiro, line 168: You know what would be nice? Instead of resetting this to zero, you could subtract 1 instead. Then, if this condition is transient, it wouldn't be triggering on/off.

Maybe something like:

```
self._count -= 1 if self._count > 0 else 0
```

python/lsst/ts/watcher/base_hexapod_overcurrent_rule.py
```diff
@@ -0,0 +1,190 @@
...
+
+        # Check the current and update the count
+        # In the normal operation, the motor current can be higher than the
+        # threshold. Therefore, we only care about the condition that we always
+        # have the motor current to be higher than the threshold all the time
+        # when there is no movement.
+        if self._is_enabled_and_stationary() and any(
+            [current >= self.config.threshold_current for current in data.motorCurrent]
+        ):
+            self._count += 1
```
tribeiro, line 166: So, python won't limit the maximum size of an integer but this could really grow up indefinitely if unbounded like this. I suggest you add a maximum allowed value here. Maybe define a `MAX_COUNT` constant and then only add up if `self._count` is smaller than that?

something like:

```
MAX_COUNT = 2147483647  # This is actually the max value of int32.

...
            self._count += 1 if self._count < MAX_COUNT else 0
```

tribeiro (commented):
This looks good, however, I think you can get away with not having to subclass a base class by making the Hexapod index/name a configuration parameters.

Also, check my inline comments about ensuring the count is bounded and about subtracting it to allow the rule to more gracefully transition from one alarm level to another.

python/lsst/ts/watcher/base_hexapod_overcurrent_rule.py
teweitsai: Updated in the commit: [c789acd](https://github.com/lsst-ts/ts_watcher/pull/94/commits/c789acde31876242324909796fb547c8ff2cf2f0). Thanks!

python/lsst/ts/watcher/base_hexapod_overcurrent_rule.py
teweitsai: Updated in the commit: [c789acd](https://github.com/lsst-ts/ts_watcher/pull/94/commits/c789acde31876242324909796fb547c8ff2cf2f0). Thanks!

python/lsst/ts/watcher/base_hexapod_overcurrent_rule.py
teweitsai: Updated in the commit: [c789acd](https://github.com/lsst-ts/ts_watcher/pull/94/commits/c789acde31876242324909796fb547c8ff2cf2f0). Thanks!

python/lsst/ts/watcher/base_hexapod_overcurrent_rule.py
teweitsai: Updated in the commit: [c789acd](https://github.com/lsst-ts/ts_watcher/pull/94/commits/c789acde31876242324909796fb547c8ff2cf2f0). Thanks!

python/lsst/ts/watcher/base_hexapod_overcurrent_rule.py
teweitsai: Not sure I understand the use of SAL index here. Do you mean to use the `SalIndex` in the configuration parameter to say we want to have the M2 hexapod alarm or camera hexapod alarm? But I thought we should have two alarms for two hexapods together because:

1. Both of them would have the overcurrent.
2. To be able to mute the individual alarm.

Please help to clarify more. Maybe I do not understand your meaning.

Thanks!

teweitsai: Updated in the commit: [70904dd](https://github.com/lsst-ts/ts_watcher/pull/94/commits/70904dd567affd77fdcbd259644d369adf303a71). Thanks!

tribeiro (approved):
Looks good! Thanks!

PR #95, opened 2025-02-17
DM-48716: Investigate memory leak in the Watcher.

tribeiro:
In addition to fixing the memory leak, improve unit test reliability and remove references to idl.

### DIFF
--- a/tests/test_remote_wrapper.py
+++ b/tests/test_remote_wrapper.py
@@ -28,7 +28,9 @@ from lsst.ts import salobj, utils, watcher
 STD_TIMEOUT = 5  # Max time to send/receive a topic (seconds)
 LONG_TIMEOUT = 60  # Max Remote startup time (seconds)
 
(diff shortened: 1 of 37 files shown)


tests/test_remote_wrapper.py
```diff
@@ -28,7 +28,7 @@
 STD_TIMEOUT = 5  # Max time to send/receive a topic (seconds)
 LONG_TIMEOUT = 60  # Max Remote startup time (seconds)
 
-index_gen = utils.index_generator()
+index_gen = utils.index_generator(imin=1234)
```
teweitsai, line 31: Please help to comment the fix here. I do not understand why the `imin=1234` fixes the issue. Thanks!

teweitsai (approved):
Looks good.

teweitsai:
For the remove of `idl`, I think you might want to modify the `conda/meta.yaml` as well.

PR #104, opened 2025-05-19
Tickets/dm 50911

iglesu:
Watcher alarm list documentation.

It uses the existing `automodapi` directive already in `developer_guide.rst`, which automatically generates documentation for all the classes in the `lsst.ts.watcher.rules` module. It essentially creates a list of alarms with descriptions pulled directly from their docstrings.

.. automodapi:: lsst.ts.watcher.rules

Example:
https://ts-watcher.lsst.io/builds/361/index.html#list-of-watcher-alarms

### DIFF
--- a/doc/alarm_list.rst
+++ b/doc/alarm_list.rst
@@ -1,0 +1,11 @@
+.. _alarm-list:
+
+List of Watcher Alarms
+======================
+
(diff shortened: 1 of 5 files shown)


tribeiro:
Hi @iglesu , thank you very much for this! I think I like option 1 better, it looks really nice!

tribeiro (approved):

PR #109, opened 2025-06-04
Tickets/osw 337

wvreeven:


python/lsst/ts/watcher/rules/mtdome_subsystem_enabled.py
```diff
@@ -99,11 +149,24 @@ def compute_alarm_severity(
...
+
+        if not self.may_raise:
+            return NoneNoReason
+
         return (
             NoneNoReason
-            if not data.faultCode
+            if not self.subsystem_state or not self.subsystem_state.fault_code
             else (
                 AlarmSeverity.CRITICAL,
```
tribeiro, line 168: you added `alarm_severity` to the schema but you are not using anywhere that I can see. Was the idea to have it here?

python/lsst/ts/watcher/rules/mtdome_subsystem_enabled.py
```diff
@@ -99,11 +149,24 @@ def compute_alarm_severity(
...
-                f"Dome az rotation state {EnabledState(data.state)!r}: {data.faultCode}.",
+                f"Dome az rotation state {self.subsystem_state.state!r}: "
```
tribeiro, line 169: You generalized the alarm to any subsystem, you probably need to update the text here to log the appropriate subsystem no? Maybe the configuration will require a provided subsystem name?

python/lsst/ts/watcher/rules/mtdome_subsystem_enabled.py
```diff
@@ -49,21 +59,61 @@ def __init__(self, config, log=None):
...
+    event_name:
+        description: The name of the event to monitor.
+        enum: {events}
+    csc_state:
+        description: The state(s) of the CSC for which the alarm is active.
+        type: array
+        minItems: 1
+        items:
+            enum: [{state_str}]
+    severity:
```
tribeiro, line 106: You included the severity here but I don't think you are using it anywhere.

python/lsst/ts/watcher/rules/mtdome_subsystem_enabled.py
```diff
@@ -99,11 +149,24 @@ def compute_alarm_severity(
...
+            self.may_raise = State(data.summaryState).name in self.config.csc_state
+            self.log.debug(f"{State(data.summaryState).name=!r}, {self.may_raise=}")
+
+        if not self.may_raise:
+            return NoneNoReason
+
         return (
             NoneNoReason
-            if not data.faultCode
+            if not self.subsystem_state or not self.subsystem_state.fault_code
```
tribeiro, line 166: here you are checking `not self.subsystem_state` I think you are trying to handle the condition when `subsystem_state` is `None`, right? if so, I think it is best to do `self.subsystem_state is not None` instead.

python/lsst/ts/watcher/rules/mtdome_subsystem_enabled.py
wvreeven: Thanks for catching that.

python/lsst/ts/watcher/rules/mtdome_subsystem_enabled.py
wvreeven: Good catch and suggestion.

PR #111, opened 2025-07-21
Fix version module import.

wvreeven:


pyproject.toml
```diff
@@ -1,32 +1,26 @@
 [build-system]
-requires = [ "setuptools", "setuptools_scm" ]
+requires = ["setuptools", "setuptools_scm"]
```
mareuter, line 2: Don't know if we want to be pedantic about the versions that the docs describe:
`requires = ["setuptools>=64", "setuptools-scm>=8"]`

mareuter (approved):
One small thing to consider, but otherwise this look good.

PR #114, opened 2025-07-25
Add MTAirCompressurePressure rule

pkubanek:


tribeiro (changes requested):
Sorry, can I get a clarification on why we are removing the power_outage rule?

pkubanek:
> Sorry, can I get a clarification on why we are removing the power_outage rule?

That was a mistake. The power_outage rule was changed to using ESS:303 CSC.

PR #115, opened 2025-08-06
Add a generic boolean rule for alarm telemetry/event items.

wvreeven:


python/lsst/ts/watcher/rules/generic_boolean.py
```diff
@@ -0,0 +1,162 @@
...
+
+        alarm_reasons = []
+        for alarm_item in self.config.alarm_items:
+            item_value = getattr(data, alarm_item, False)
+            if item_value is True:
+                severity = AlarmSeverity.CRITICAL
+                alarm_reasons.append(f"{alarm_item} is {item_value}")
+
+        if len(alarm_reasons) > 0:
+            reason = (
```
tribeiro, line 158: I would drop the `remote_name` and `remote_index` from the reason. This information is already in the alarm name.

I would just do something like:

```
reason = "\n".join(alarm_reasons)
```

python/lsst/ts/watcher/rules/generic_boolean.py
```diff
@@ -0,0 +1,162 @@
...
+        Notes
+        -----
+        You may return `NoneNoReason` if the alarm state is ``NONE``.
+        """
+        severity, reason = NoneNoReason
+
+        alarm_reasons = []
+        for alarm_item in self.config.alarm_items:
+            item_value = getattr(data, alarm_item, False)
+            if item_value is True:
```
tribeiro, line 153: what if the boolean value is reversed? meaning True is ok and False is bad?

Also, note that this is not the correct way to check for boolean, you should just do:

```
if item_value:
```

e.g. https://developer.lsst.io/python/style.html#is-and-is-not-should-only-be-used-for-determining-if-two-variables-point-to-same-object

but I think you should make this configurable and than do something like:

```
item_value = getattr(data, alarm_item, self.expected_value)
if item_value != self.expected_value:
    severity = AlarmSeverity.CRITICAL
    alarm_reasons.append(f"{alarm_item} is {item_value}; expected {self.expected_value}.")
```

python/lsst/ts/watcher/rules/generic_boolean.py
```diff
@@ -0,0 +1,162 @@
...
+                severity = AlarmSeverity.CRITICAL
```
tribeiro, line 154: You included the severity in the configuration but here you are hard-coding it to `CRITICAL`.

tribeiro (commented):
Thanks for adding this alarm, I think it is going to be really useful. I do have some comments I hope you can address before merging it.

PR #131, opened 2026-04-13
Tickets/osw 2065

wvreeven:


python/lsst/ts/watcher/alarm_rule_runner.py
```diff
@@ -0,0 +1,309 @@
...
+            except asyncio.CancelledError:
+                self._should_produce_heartbeats = False
+            except Exception:
+                self.log.exception("Heartbeat output failed.")
+                self._should_produce_heartbeats = False
+
+    async def _potentially_wait_for_task_done(self, task: asyncio.Future, should: bool) -> None:
+        if not task.done():
+            self.log.debug(f"Waiting for task {task} to be done.")
+            should = False  # noqa
```
tribeiro: Do you need to do this? Doesn't seem like you are using `should` anywhere else. This is also creating a local variable shadowing the input parameter.

python/lsst/ts/watcher/alarm_rule_runner.py
```diff
@@ -0,0 +1,309 @@
...
+            try:
+                await asyncio.sleep(HEARTBEAT_INTERVAL)
+                await self.evt_heartbeat.write()
+            except asyncio.CancelledError:
+                self._should_produce_heartbeats = False
+            except Exception:
+                self.log.exception("Heartbeat output failed.")
+                self._should_produce_heartbeats = False
+
+    async def _potentially_wait_for_task_done(self, task: asyncio.Future, should: bool) -> None:
```
tribeiro: It took me some time to understand what this method is supposed to do. Let me see if I got this right, 

You are trying to make it so you will wait for the task to finish but it does not care about the result of the task (regardless if it is an exception or not) and you wan to cancel the task if it timeout. The way you did it was by polling on the task `done` status, and just ignoring the actual result of the task. 

I would suggest you add a docstring to explain what this method is supposed to do for our future selves.

Also, I don't think you are using the `should` input parameter and I am not sure I understand what it was supposed to do.

python/lsst/ts/watcher/alarm_rule_runner.py
```diff
@@ -0,0 +1,309 @@
...
+        self._should_produce_heartbeats = True
+        self._heartbeat_task = asyncio.create_task(self._heartbeat_loop())
+
+    async def _run(self) -> None:
+        """Run the alarm rule."""
+        await self._set_state(AlarmRuleState.RUNNING)
+        self._should_be_running = True
+        await self.model.enable()
+
+        while self._should_be_running:
```
tribeiro: Maybe instead of doing this, use a `Future` and wait for it here? then you could do something like

```
self._should_be_running = asyncio.Future()

await self._should_be_running
```

and, wherever you would set `self._should_be_running = False`, you would do `self._should_be_running.set_result(None)` to indicate it is done.

python/lsst/ts/watcher/alarm_rule_runner.py
```diff
@@ -0,0 +1,309 @@
...
+            self.model.disable()
+            await self.model.close()
+            self.model = None
+
+        await self._potentially_wait_for_task_done(self._run_task, self._should_be_running)
+        await self._potentially_wait_for_task_done(self._heartbeat_task, self._should_produce_heartbeats)
+
+    async def _stop(self) -> None:
+        """Stop the alarm rule and all tasks."""
+        await self._set_state(AlarmRuleState.STOPPING)
```
tribeiro, line 132: You could use some syntactic sugar here to improve the "stopping" -> "stopped" operation by using an `asynccontextmanager`. Something like:

```
@contextlib.asynccontextmanager
async def stopping():
    await self._set_state(AlarmRuleState.STOPPING)
    yield
    await self._set_state(AlarmRuleState.STOPPED)
```

Then you could do:

```
async with self.stopping():
    await self.close_tasks()
```

python/lsst/ts/watcher/alarm_rule_runner.py
```diff
@@ -0,0 +1,309 @@
...
+        ------
+        salobj.ExpectedError
+            If ``self.state.state`` is not
+            `lsst.ts.xml.enums.AlarmRule.AlarmRuleState.UNCONFIGURED`.
+        """
+        self.log.debug("do_configure")
+
+        expected_state = AlarmRuleState.UNCONFIGURED
+        if self.state != expected_state:
+            raise RuntimeError(f"AlarmRule state is not {expected_state.name}.")
```
tribeiro: Consider rephrasing it to include the current state in the error message. Something like:

```
raise RuntimeError(
    f"Invalid AlarmRule state {self.state.name}; "
    f"expected {expected_state.name}."
)
```

Also, you might want to consider creating a specific exception for this, e.g.:

```
class UnexpectedAlarmRuleStateError(Exception):
    pass
```

python/lsst/ts/watcher/alarm_rule_runner.py
```diff
@@ -0,0 +1,309 @@
...
+        ------
+        salobj.ExpectedError
+            If ``self.state.state`` is not
+            `lsst.ts.xml.enums.Script.AlarmRuleState.CONFIGURED`.
+        """
+        self.log.debug("do_run")
+
+        expected_state = AlarmRuleState.CONFIGURED
+        if self.state != expected_state:
+            raise RuntimeError(f"AlarmRule state is not {expected_state.name}.")
```
tribeiro: Same as above, consider rephrasing to add the current state and using a custom exception (e.g. `UnexpectedAlarmRuleStateError`).

python/lsst/ts/watcher/alarm_rule_runner.py
```diff
@@ -0,0 +1,309 @@
...
+
+        try:
+            self._run_task = asyncio.create_task(self._run())
+            await self._run_task
+        except asyncio.CancelledError:
+            # Deliberately ignore.
+            pass
+        except BaseException as e:
+            if not isinstance(e, salobj.ExpectedError):
```
tribeiro, line 239: You really only want to log the exception if it is an `ExpectedError`? might be worth always logging it no?

python/lsst/ts/watcher/alarm_rule_runner.py
```diff
@@ -0,0 +1,309 @@
...
+                self.log.exception("Error in run.")
+            await self._set_state(AlarmRuleState.FAILING)
```
tribeiro: You might want to use the context manager I suggested for stopping here as well...

tribeiro (approved):
This is really nice. I have a few inline comments I hope you will consider before merging but otherwise looks good!

python/lsst/ts/watcher/alarm_rule_runner.py
wvreeven: I think you missed the `not` in the if statement. The error is always logged unless it is an ExpectedError.

python/lsst/ts/watcher/alarm_rule_runner.py
wvreeven: That is a really smart idea. Thanks! I really was unhappy with that piece of code.

PR #132, opened 2026-05-05
Tickets/osw 2208

mxturri:


python/lsst/ts/watcher/rules/mt_camera_alert.py
```diff
@@ -49,6 +50,12 @@ class MTCameraAlert(BaseRule):
     def __init__(self, config, log=None):
         rule_name = "MTCameraAlert"
         remote_name = "MTCamera"
+
+        # The CCS Alert ID as provided by the configuration.
+        # This is the unique identifier used to differentiate the
+        # rules and associate each of them to the corresponding
```
wvreeven: I think it is `associate with` and not `associate to`.

python/lsst/ts/watcher/rules/mt_camera_alert.py
```diff
@@ -49,6 +50,12 @@ class MTCameraAlert(BaseRule):
     def __init__(self, config, log=None):
         rule_name = "MTCameraAlert"
         remote_name = "MTCamera"
```
wvreeven, line 52: This variable is no longer used so please remove it. I am surprised that this wasn't detected by mypy or ruff.

tests/rules/test_mt_camera_alert.py
```diff
@@ -32,76 +34,180 @@
...
-
-            alert_id = "TestId"
-            description = "A test alert"
-            cause = "Test cause"
-            origin = "lsstcam"
-            additional_info = "AdditionalInfo"
-
-            for is_cleared in [False, True]:
+        async with salobj.Controller(name="MTCamera", index=0) as controller:
+            async with watcher.Model(domain=controller.domain, config=watcher_config) as model:
```
wvreeven: There is no reason to separate these statements. The original code worked fine so please revert to
```
async with (
            salobj.Controller(name=self.remote_name, index=0) as controller,
            watcher.Model(domain=controller.domain, config=watcher_config) as model,
        ):
```

tests/rules/test_mt_camera_alert.py
```diff
@@ -32,76 +34,180 @@
...
+              configs:
+              - alertId: {alert_id1}
+              - alertId: {alert_id2}
+            escalation: []
+            """
+        )
+
+        watcher_config = types.SimpleNamespace(**watcher_config_dict)
+        async with salobj.Controller(name="MTCamera", index=0) as controller:
+            async with watcher.Model(domain=controller.domain, config=watcher_config) as model:
```
wvreeven: There is no reason to separate these statements. Please change to
```
async with (
            salobj.Controller(name=self.remote_name, index=0) as controller,
            watcher.Model(domain=controller.domain, config=watcher_config) as model,
        ):
```

wvreeven (changes requested):
Good job! However, I left a few comments for you. Please review them and make the requested changes. Then please re-request my review in the PR.

python/lsst/ts/watcher/rules/mt_camera_alert.py
mxturri: I see that variable "remote_name" is used further down the code. So I didn't do anything about this suggestion.

tests/rules/test_mt_camera_alert.py
mxturri: Thanks! I have modified both tests to have a single statement.
I had originally split them in an attempt to see if this might help with the strange error failure I was getting. I checked other tests and I noticed that some were using a single async block, while other were using two of them split apart.

python/lsst/ts/watcher/rules/mt_camera_alert.py
wvreeven: My bad! Thanks for checking.

wvreeven (approved):
Looks good to me!

PR #135, opened 2026-07-08
Make DewPointDepression handle non-ESS telemetry.

wvreeven:


edennihy (approved):
Approved. Tested on BTS on 2026-07-08.

PR #137, opened 2026-07-13
Fix bug where no telemetry seen was reported while telemetry seen.

wvreeven:


doc/news/OSW-2688.bugfix.rst
edennihy, line 1: Really picky comment, but this should probably  `OSW-2688.bugfix.rst` instead of .perf.rst

edennihy (approved):
Approved.

PR #138, opened 2026-08-05
Tickets/osw 2243

wvreeven:


python/lsst/ts/watcher/alarm_rule_runner.py
```diff
@@ -333,10 +389,148 @@ async def output_alarm(self, alarm):
...
-    async def amain(cls, **kwargs: typing.Any) -> None:
-        raise NotImplementedError()
+    async def amain(cls) -> None:
+        """Run an AlarmRuleRunner instance from the command line."""
+        runner = cls.make_from_cmd_line()
+
+        try:
+            await runner.done_task
+            await runner.close()
+        except Exception as e:
```
tribeiro: Do you also want to handle `CancelledError`? Remember that some exceptions are now subclassing `BaseException` instead of `Exception` so this is not catching everything... :(

python/lsst/ts/watcher/watcher_csc.py
```diff
@@ -83,17 +100,49 @@ def __init__(self, config_dir=None, initial_state=salobj.State.STANDBY, override
...
+            await alarm_rule_info.remote.cmd_stop.set_start()
+            await alarm_rule_info.remote.close()
+            self.log.debug(f"Remote[AlarmRule:{index}] is stopped.")
+
+            process = alarm_rule_info.process
+            self.log.debug(
+                f"Waiting for alarm rule process {process.pid} for rule {alarm_rule_info.classname} to exit."
+            )
+            while process.returncode is None:
+                await asyncio.sleep(0.1)
```
tribeiro: Isn't there a task you can wait instead?

python/lsst/ts/watcher/watcher_csc.py
```diff
@@ -130,206 +179,261 @@ async def configure(self, config):
...
+            process = await asyncio.create_subprocess_exec(
+                SCRIPT_NAME, rule["classname"], str(index), stdin=asyncio.subprocess.PIPE
+            )
 
-        Raises
-        ------
-        RuntimeError
-            If pre-conditions are not met (escalation is not attempted):
+            self.log.debug(f"Creating Remote[AlarmRule:{index}] for rule {rule['classname']}.")
+            alarm_rule_remote = salobj.Remote(domain=self.domain, name="AlarmRule", index=index)
```
tribeiro: So, on potential problem here is that you are creating a new remote for each rule, this means a lot of duplication of resources. You could instead have a single `AlarmRule` remote and have a callback to offload the information depending on the index. However, I worry that you might have driven yourself into a corner here by passing the remote into `AlarmRuleInfo`.  I think you would have to refactor this to have methods in `AlarmRuleInfo` that needs to be called to update the information, then the callback would farm off the data as it receives it.

tribeiro (commented):
These changes are pretty massive, which makes it hard to do a more in-depth review. One thing that makes it even more challenging is that the bulk of the changes are all done in a single commit d475413. If you had split the contributions in several commits, each commit with a set of incremental changes, it would have made it more easy to review 😅.

So, one thing that worries me a bit about this implementation is that you are creating one individual `Remote` for each `AlarmRule` instance. This can be quite resource intensive. You could make this a bit better, though slightly more challenging to manage, by creating a single `Remote` with `index=0`. For an indexed component `index=0` means, read data from all instances/indices. Then, you could have callbacks that read the index and call method in the `AlarmRuleInfo` to update the information. You can also use the same remote to send commands to individual instances by passing `salIndex` as an argument to `set_start`, e.g.:

```
await self.alarm_rule_remote.cmd_configure.set_start(config=alarm_rule_config_str, salIndex=index, timeout=STD_TIMEOUT)
```

One general comment, I noticed the conda recipe is not passing along the `salobj_version` and `xml_version` flags to the runtime dependency. You might want to fix that, especially in this case where it will no longer support older versions of salobj.

python/lsst/ts/watcher/watcher_csc.py
wvreeven: As far as I can tell from [the documentaiton](https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.subprocess.Process), no.

tribeiro (commented):
Nice work! I do have a few more inline comments I hope you will consider. Also, I noticed that a few commands have no specified timeouts. I think you might want to ensure that all commands have timeouts to avoid locking indefinitely in case of a network glitch or some kind of bug.

python/lsst/ts/watcher/alarm_rule_runner.py
```diff
@@ -282,7 +285,10 @@ async def do_stop(self, data: type_hints.BaseMsgType) -> None:
         This is usually called when the Watcher goes to DISABLED state.
         """
         self.log.debug("do_stop")
-        asyncio.create_task(self._stop())
+        if not self._stop_task.done():
+            self.log.warning(f"{self.rule_name}:{self.salinfo.index} already stopping.")
```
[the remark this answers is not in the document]
wvreeven: Added the `return` that I forgot.

python/lsst/ts/watcher/watcher_csc.py
```diff
@@ -83,17 +100,63 @@ def __init__(self, config_dir=None, initial_state=salobj.State.STANDBY, override
...
+            process = alarm_rule_info.process
+            self.log.debug(
+                f"Waiting for alarm rule process {process.pid} for rule {alarm_rule_info.classname} to exit."
+            )
+            while process.returncode is None:
+                await asyncio.sleep(0.1)
+            self.log.debug(f"Alarm rule process {process.pid} exited with code {process.returncode}.")
+
+        self.log.debug("Stopping AlarmRuleRemote.")
+        await self.alarm_rule_remote.close()
```
tribeiro: You are still closing the remote here... You probably don't want to do that anymore

python/lsst/ts/watcher/watcher_csc.py
```diff
@@ -130,206 +192,265 @@ async def configure(self, config):
...
-        await self.model.start_task
-
-    async def escalate_alarm(self, alarm):
-        """Escalate an alarm by creating a SquadCast incident.
-
-        Store the ID of the alert in alarm.escalation_id.
-        If the attempt fails, store an error message that begins with
-        "Failed: " in alarm.escalation_id.
+        self.log.debug("Waiting for AlarmRuleRemote start_task.")
+        await self.alarm_rule_remote.start_task
```
tribeiro: I don't think you need this anymore...

tribeiro (approved):
spotted a few more things in the last run down... I think it would be good to test this on BTS before merging, just to make sure we can spot any last minute bug etc.... 

But, in any case, before merging take a look at the pair of comments I left. At least one of them is pretty important (you need to stop closing the remote). The other one is a no-op (waiting for start_task).

